PR #16, opened 2020-02-14
Upgrade to Rector 0.7

shaal:


shaal:
@damontgomery the last thing that prevent me from finishing the upgrade to Rector 0.7.11, is an error message:
```
[ERROR] Could not process "drupal-rector/rector_examples/src/EntityManagerStaticUpdated.php" file,
         due to:                                                                                   
         "Pick more specific node than "PhpParser\Node\Expr\MethodCall"".  
```
I believe this is the line that causes the problem, but I don't know what would be the fix for that:
https://github.com/palantirnet/drupal-rector/blob/master/src/Rector/Deprecation/EntityManagerRector.php#L51

TomasVotruba:
The break is caused by change of `getName()` method. It reported false positives before on chains static/method calls, so now it always needs the `->name` part too:

```diff
 /** @var MethodCall $node */
 $node = ...;
-$this->getName($node)
+$this->getName($node->name)
```

```diff
 /** @var StaticCall $node */
 $node = ...;
-$this->getName($node)
+$this->getName($node->name)
```

Just update every `Rector` rule and `getName()/isName()/isNames()` method and it should work.

damontgomery:
Thanks, I made this change in `master`. It looks like it works in 0.6 and should work in 0.7 (not tested).

TomasVotruba (approved):

shaal:
Test is working as intended:
https://github.com/palantirnet/drupal-rector/runs/595807276?check_suite_focus=true#step:8:4

.github/workflows/drupal_rector.yaml
```diff
@@ -0,0 +1,44 @@
...
+            -   uses: actions/checkout@v2
+            -   uses: shivammathur/setup-php@v2
+                with:
+                    php-version: 7.3
+                    coverage: none # disable xdebug, pcov
+                    extensions: "intl"
+
+            # Run standalone install in non-root package, ref https://github.com/rectorphp/rector/issues/732
+            -   run: |
+                    # Install latest Drupal
```
damontgomery: When I read `Install latest Drupal` I interpret this as meaning that we install a working site with a database, etc. I understand from reading this that we are downloading the latest Drupal codebase, but I was thrown off.

We can update this later if we come up with better comments.

.github/workflows/drupal_rector.yaml
```diff
@@ -0,0 +1,44 @@
...
+                    composer create-project drupal/recommended-project drupal --dev --no-progress
+
+            -   run: |
+                    # Install drupal-rector (based on rector 0.7.x)
+                    cd drupal
+                    composer require palantirnet/drupal-rector:dev-rector-0.7 --no-progress
```
damontgomery: Does this need to be updated to remove the branch name?

Do you want to do that after merging?

composer.json
```diff
@@ -11,7 +11,7 @@
     ],
     "require": {
         "jawira/case-converter": "^1.1",
-        "rector/rector-prefixed": "0.6.13"
+        "rector/rector-prefixed": "~0.7.16"
```
damontgomery: Should we use `^0.7.16`? This will work up to but not including `0.8`. The `~` seems to be a little different before and after `1.0` releases.

damontgomery (commented):
Thanks for this, @shaal.

I left a few small comments about composer versions and comments. Can you take a look?

.github/workflows/drupal_rector.yaml
shaal: Yes, after merging we would change this

composer.json
TomasVotruba: I recommend using this, because the sooner you discover BC break, the sooner I can help you fix it.

TomasVotruba:
:partying_face: :tada:

PR #18, opened 2020-03-17
Rector rule for deprecated GetMock()

nerdstein:
https://www.drupal.org/node/2907725


rector_examples/src/GetMock.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\rector_examples\GetMock.
+ */
+
+namespace Drupal\rector_examples;
+
+class GetMock {
```
damontgomery: If possible, it would be great to have this example service extend one of the base classes where this would be found in Drupal. The examples on Drupal.org for the change record indicate that this would be called with `$this->getMock(string $class)` when we have a base class of `UnitTestCase`, `KernelTestBase`, or `BrowserTestBase`. I think if we create examples of each that would be good.

Can we use something like `$this->getMock('my_class')` in the example?

I'm pretty sure on the string as argument, but not 100%.

The reason for the base classes is that we may be able to check the class type in Rector (as a nice-to-have) and it would have to probably be one of those three classes.

src/Rector/Deprecation/GetMockRector.php
```diff
@@ -0,0 +1,77 @@
...
+     * @inheritdoc
+     */
+    public function refactor(Node $node): ?Node
+    {
+        /** @var Node\Stmt\Expression $node */
+        $expr = $node->expr;
+        if ($expr instanceof Node\Expr\MethodCall && 'getMock' === (string) $expr->name) {
+
+            // Get the calling variable.
+            $var = (string) $expr->var->name;
```
damontgomery: It doesn't look like you've used this.

damontgomery (commented):
Adam, thanks so much. This is awesome!

I'm still trying to understand when to use each type of node. The statement type here seems weird to me. This one seems to be a better fit, `Node\Expr\MethodCall`. You are checking that in the refactor code, so maybe using it in the `getNodeTypes()` would also be good.

I was talking with Ofer before about some base classes, and I think this would be a great base class where we could set a `private deprecated_method_name = getMock;` and `private new_method_name = createMock;` on a class that inherits most of this functionality.

If you want to give that a shot, that would be awesome, but I'm also cool to have people take a look on Saturday during the code sprint.

I can also see a case for checking class types and the use of `$this` as the first part of the method call, but that's all nice to have. If you are interested, maybe add some 'TODOs' in the code for now.

Sorry we won't get to hang out in person this year. Thanks again and TTYL!

damontgomery:
@nerdstein, I resolved a merge conflict based on a recent update.

nerdstein:
Need to revisit this -- https://github.com/rectorphp/rector/issues/1941

deprecation-index.yml
```diff
@@ -40,3 +40,8 @@
     - db_query_updated.php
     - DBQueryStatic.php
     - DBQueryStaticUpdated.php
+'getMock()':
```
damontgomery: @shaal, let's change this to `BrowserTestBase::getMock()` and update the rector name below.

rector_examples/test/src/Functional/BrowserTestBaseGetMock.php
```diff
@@ -0,0 +1,23 @@
...
+namespace Drupal\Tests\rector_examples\Functional;
+
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Tests\BrowserTestBase;
+
+class BrowserTestBaseGetMock extends BrowserTestBase {
+
+  /**
+   *
+   */
```
damontgomery, line 12: @shaal , can you add comments for these examples?

deprecation-index.yml
shaal: 👍

rector_examples/test/src/Functional/BrowserTestBaseGetMock.php
shaal: 👍

PR #24, opened 2020-03-21
db_query

damontgomery:


damontgomery:
@nerdstein, I've included some updates here to parse in-line arguments. I haven't looked at evaluating the variables here, but I added a TODO. Hopefully this helps you proceed if you want to take a look.

I think this also does a really good job so far, so if we want to merge as-is, I'm good with that too.

nerdstein:
Variable eval here: https://github.com/rectorphp/rector/issues/1941

damontgomery:
@nerdstein, I think Ofer reached out to you about recent resourcing. I'm going to start working on this and may merge it in with basic support and a list of TODOs.

I wanted to thank you again for your work and don't want you to feel like we're excluding you. :)

PR #26, opened 2020-04-08
Capitalize DB in filenames (applied by `git mv -f`)

shaal:
After commit b719bdd, the filenames capitalization did not get into git.

That caused an error when running `composer install`:
```
Deprecation Notice: Class DrupalRector\Rector\Deprecation\DBQueryRector located in
./src/Rector/Deprecation/DbQueryRector.php does not comply with psr-4 autoloading standard.
It will not autoload anymore in Composer v2.0. in
/Users/shaal/code/palantir/drupal-rector-sandbox/vendor/composer/composer/src/Composer/Autoload/ClassMapGenerator.php:201
```

By using `git mv -f` I was able to get the new file name capitalization update into git.


damontgomery:
Thanks for fixing this Ofer.

PR #30, opened 2020-04-10
Add rector for deprecated file_prepare_directory() function.

Kbentham:
Add rector file and examples.

### DIFF
--- a/src/Rector/Deprecation/DrupalSetMessageRector.php
+++ b/src/Rector/Deprecation/DrupalSetMessageRector.php
@@ -70,7 +70,7 @@ CODE_AFTER
             } else {
                 // TODO: Add the messanger trait to a class that doesn't have it.
                 // For now, we are using a static call.
-                $messenger_service = new Node\Expr\StaticCall(new Node\Name\FullyQualified('Drupal'), 'messenger');
+                $messenger_service = new Node\Expr\StaticCall(new Node\Name\FullyQualified('Drupal'), 'file_system');
             }
 
             $method_name = 'addStatus';
--- a/src/Rector/Deprecation/DrupalSetMessageRector.php
+++ b/src/Rector/Deprecation/DrupalSetMessageRector.php
@@ -70,7 +70,7 @@ CODE_AFTER
             } else {
                 // TODO: Add the messanger trait to a class that doesn't have it.
                 // For now, we are using a static call.
-                $messenger_service = new Node\Expr\StaticCall(new Node\Name\FullyQualified('Drupal'), 'file_system');
+                $messenger_service = new Node\Expr\StaticCall(new Node\Name\FullyQualified('Drupal'), 'messenger');
             }
 
             $method_name = 'addStatus';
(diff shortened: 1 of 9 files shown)


src/Rector/Deprecation/DrupalSetMessageRector.php
```diff
@@ -70,7 +70,7 @@ public function refactor(Node $node): ?Node
             } else {
                 // TODO: Add the messanger trait to a class that doesn't have it.
                 // For now, we are using a static call.
-                $messenger_service = new Node\Expr\StaticCall(new Node\Name\FullyQualified('Drupal'), 'messenger');
+                $messenger_service = new Node\Expr\StaticCall(new Node\Name\FullyQualified('Drupal'), 'file_system');
```
damontgomery: @Kbentham , is this a copy & paste error?

src/Rector/Deprecation/FilePrepareDirectoryRector.php
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * What is covered:
+ * - Static replacement
+ *
+ * Improvement opportunities
+ * - Dependency injection
+ */
+final class FilePrepareDirectoryRector extends AbstractRector
+{
+    use TraitsByClassHelperTrait;
```
damontgomery: This can be removed since we aren't checking traits on classes.

src/Rector/Deprecation/FilePrepareDirectoryRector.php
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+     * @inheritdoc
+     */
+    public function getDefinition(): RectorDefinition
+    {
+        return new RectorDefinition('Fixes deprecated drupal_set_message() calls',[
```
damontgomery: Can you update the comment here where it says `drupal_set_message()`?

src/Rector/Deprecation/FilePrepareDirectoryRector.php
```diff
@@ -0,0 +1,76 @@
...
+        ];
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function refactor(Node $node): ?Node
+    {
+        /** @var Node\Expr\FuncCall $node */
+        if ($node->name instanceof Node\Name && 'file_prepare_directory' === (string) $node->name) {
```
damontgomery: @Kbentham, this works, but I think it might be good to use `$this->getName($node) === 'file_prepare_directory'` which might be easier to read. Do you want to try that?

src/Rector/Deprecation/FilePrepareDirectoryRector.php
```diff
@@ -0,0 +1,76 @@
...
+
+            // This creates a service call like `\Drupal::service('file_system').
+            // TODO use dependency injection.
```
damontgomery, line 63: Thanks for the comment!

damontgomery (commented):
@Kbentham, Thanks for this!

I left a few comments. I think only the one on the DrupalSetMessageRector is important. Can you take a look at them?

Thanks for all the examples and comments!

damontgomery:
Thanks, @Kbentham!

PR #36, opened 2020-04-13
FILE_CREATE_DIRECTORY rector.

damontgomery:


rector_examples/file_create_directory_updated.php
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * This demonstrates the updated deprecated static calls that might be called from procedural code like `.module` files.
+ */
+
+/**
+ * A simple example.
+ */
+function simple_example() {
+  $x = \Drupal\Core\File\FileSystemInterface::CREATE_DIRECTORY;
```
agentrickard, line 13: Would this need the full class path?

damontgomery: You can use the full path or a `use` statement. The rector puts the full path since it always works and we can provide the use statement later as an enhancement.

I provided examples of both here to illustrate how it could work. We can discuss how we want to handle examples if that would be helpful.

agentrickard: I'm not sure. It's just odd since there _is_ a `use` statement here. I would strive to use shortnames whenever `use` is invoked.

damontgomery: I don't know the best balance of showing examples.

I agree that use statements and short names should be used when possible in regular development. This example shows how rector is currently updating the code. It then shows an alternative of how to use a use statement.

I agree it's confusing. I've tried to show a variety of "fixed" examples whether or not rector can currently fix them. drupal_set_message() has an example of this where a case that isn't yet handled is shown as fixed so that people can use it as a reference and we can use it as a target. We should probably be explicit with our "updated" examples to indicate how code is fixed by rector and separately how it "should" be fixed ideally. That's a lot more work, particularly around dependency injection, but we can revisit the examples if that is valuable.

agentrickard: I may just be reading things out of context.

PR #38, opened 2020-04-15
Add rector for deprecated drupal_render() and drupal_render_root functions

shaal:
Add rector file and examples.

### DIFF
--- a/rector_examples/drupal_render.php
+++ b/rector_examples/drupal_render.php
@@ -8,6 +8,9 @@
  * A simple example using the minimum number of arguments.
  */
 function simple_example() {
+  $elements = [
+    '#markup' => '<div>hello world</div>',
+  ];
   drupal_render($elements);
 }
 
@@ -15,7 +18,9 @@ function simple_example() {
  * An example using all of the arguments.
  */
 function using_all_arguments() {
-  $is_recursive_call = false;
+  $is_recursive_call = FALSE;
+  $elements = [
+    '#markup' => '<div>hello world</div>',
+  ];
   drupal_render($elements, $is_recursive_call);
 }
-
--- a/rector_examples/drupal_render.php
+++ b/rector_examples/drupal_render.php
@@ -1,0 +1,21 @@
+<?php
+
+/**
+ * This demonstrates the deprecated static calls that might be called from procedural code like `.module` files.
+ */
+
+/**
+ * A simple example using the minimum number of arguments.
+ */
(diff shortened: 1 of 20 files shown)


rector_examples/drupal_render.php
```diff
@@ -0,0 +1,21 @@
...
+ */
+function simple_example() {
+  drupal_render($elements);
+}
+
+/**
+ * An example using all of the arguments.
+ */
+function using_all_arguments() {
+  $is_recursive_call = false;
```
damontgomery: @shaal, Php coding standard is to use all-caps for booleans. `FALSE`.

rector_examples/drupal_render.php
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * This demonstrates the deprecated static calls that might be called from procedural code like `.module` files.
+ */
+
+/**
+ * A simple example using the minimum number of arguments.
+ */
+function simple_example() {
+  drupal_render($elements);
```
damontgomery, line 14: Can you define `$elements` here?

A simple example is
```
$elements = [
  '#markup' => '<div>hello world</div>',
];
```

rector_examples/src/DrupalRenderRoot.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+namespace Drupal\rector_examples;
+
+/**
+ * Example of static method calls from a class.
+ */
+class DrupalRenderRootStatic {
```
damontgomery: You need to name this file `DrupalRenderRootStatic` to match the class name.

rector_examples/src/DrupalRenderRootUpdated.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+namespace Drupal\rector_examples;
+
+/**
+ * Example of static method calls from a class.
+ */
+class DrupalRenderRootStatic {
```
damontgomery: You need to name the file and class `DrupalRenderRootStaticUpdated`.

damontgomery (changes requested):
Thanks, @shaal.

Overall this looks pretty good. I think the examples would be more helpful if they defined all the variables. I've left comments.

Did you miss the `drupal_render_root.php` example file?

Can you define the classes in the `services.yml` file in the example module?

damontgomery:
Thanks, @shaal. Looks good!

PR #39, opened 2020-04-16
file_scan_directory rule

shaal:
[WIP]

@damontgomery according to the change-record, https://www.drupal.org/node/3038437
We should embed this inside a new `if` statement.
Do you know how can we implement that with Rector?

### DIFF
--- a/rector_examples/file_scan_directory.php
+++ b/rector_examples/file_scan_directory.php
@@ -1,0 +1,30 @@
+<?php
+
+/**
+ * This demonstrates the deprecated static calls that might be called from procedural code like `.module` files.
+ */
+
+/**
+ * A simple example using the minimum number of arguments.
+ */
+function simple_example() {
+  $directory = '/test/directory';
+
+  file_scan_directory($directory);
+}
+
+/**
+ * An example using all of the arguments.
+ */
+function using_all_arguments() {
+  $directory = '/test/directory';
+  $mask = '/^' . DRUPAL_PHP_FUNCTION_PATTERN . '$/';
+  $options = [];
+  $options += [
+    +      'callback' => 0,
+    +      'recurse' => TRUE,
+    +      'key' => 'uri',
+    +      'min_depth' => 0,
+  ];
+  file_scan_directory($directory, $mask, $options);
+}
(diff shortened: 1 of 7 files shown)


rector_examples/file_scan_directory.php
```diff
@@ -0,0 +1,30 @@
...
+function using_all_arguments() {
+  $directory = '/test/directory';
+  $mask = '/^' . DRUPAL_PHP_FUNCTION_PATTERN . '$/';
+  $options = [];
+  $options += [
+    +      'callback' => 0,
+    +      'recurse' => TRUE,
+    +      'key' => 'uri',
+    +      'min_depth' => 0,
+  ];
```
damontgomery, line 27: Is this WIP code? I would set the options rather than merging arrays.

src/Rector/Deprecation/FileScanDirectoryRector.php
```diff
@@ -0,0 +1,47 @@
+<?php
+
+namespace DrupalRector\Rector\Deprecation;
+
+use Rector\RectorDefinition\CodeSample;
+use Rector\RectorDefinition\RectorDefinition;
+
+/**
+     * Replaces deprecated file_scan_directory() calls.
```
damontgomery: Can you fix this spacing?

src/Rector/Deprecation/FileScanDirectoryRector.php
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * See https://www.drupal.org/node/3038437 for change record.
+ *
+ * What is covered:
+ * - Static replacement
+ *
+ * Improvement opportunities
+ * - Dependency injection
```
damontgomery: Can you add a comment about adding the if() statement?

damontgomery (changes requested):
Overall this makes sense to me. Do we want to move forward without the `if` statement and come back to that later? I don't think it would be needed in most cases.

Kbentham:
I have merged in master and updated this pr to account for all of the changes and dan's feedback.

damontgomery:
Thanks @Kbentham and @shaal!

PR #41, opened 2020-04-17
Fix drupal_render_root() message

goba:


shaal:
Thank you for fixing it!

PR #43, opened 2020-04-17
feature-rectore-3115670: rule for file_unmanaged_copy

mohsin-saeed:
file_unmanaged_copy rector

### DIFF
--- a/rector_examples/file_unmanaged_copy.php
+++ b/rector_examples/file_unmanaged_copy.php
@@ -1,0 +1,10 @@
+<?php
+
+/**
+ * A simple example using the minimum required (all) number of arguments.
+ */
+function simple_example() {
+  $source = '/test/directory';
+  $destination = '/test/directory_2';
+  file_unmanaged_copy($source, $destination,FILE_CREATE_DIRECTORY);
+}
--- a/src/Rector/Deprecation/FileUnmanagedCopyRector.php
+++ b/src/Rector/Deprecation/FileUnmanagedCopyRector.php
@@ -1,0 +1,71 @@
+<?php
+
+namespace DrupalRector\Rector\Deprecation;
+
+use DrupalRector\Utility\TraitsByClassHelperTrait;
+use PhpParser\Node;
+use Rector\Rector\AbstractRector;
+use Rector\RectorDefinition\CodeSample;
+use Rector\RectorDefinition\RectorDefinition;
+
+/**
+ * Replaces deprecated file_unmanaged_copy() calls.
+ *
+ * See https://api.drupal.org/api/drupal/core%21includes%21file.inc/function/file_unmanaged_copy/8.7.x.
+ *
+ * What is covered:
+ * - File System Service used
+ *
+ * Improvement opportunities
+ * - Dependency Injection
+ */
+final class FileUnmanagedCopyRector extends AbstractRector {
+  use TraitsByClassHelperTrait;
+
+  /**
+   * @inheritdoc
+   */
+  public function getDefinition(): RectorDefinition {
+    return new RectorDefinition('Fixes deprecated file_unmanaged_copy() calls', [
+      new CodeSample(
+        <<<'CODE_BEFORE'
+file_unmanaged_copy($source, $destination, $replace);
+CODE_BEFORE
+        ,
+        <<<'CODE_AFTER'
+\Drupal::service('file_system')->copy($source, $destination, $replace);
+CODE_AFTER
+      ),
+    ]);
+  }
+
+  /**
+   * @inheritdoc
+   */
+  public function getNodeTypes(): array {
+    return [
+      Node\Expr\FuncCall::class,
+    ];
+  }
+
+  /**
+   * @inheritdoc
+   */
+  public function refactor(Node $node): ?Node {
+    if ($node->name instanceof Node\Name && 'file_unmanaged_copy' === (string) $node->name) {
+      $file_system_service = new Node\Expr\StaticCall(new Node\Name\FullyQualified('Drupal'), 'service', [new Node\Arg(new Node\Scalar\String_('file_system'))]);
+
+      $method_name = 'copy';
+
+      $method = new Node\Identifier($method_name);
+
+      $node = new Node\Expr\MethodCall($file_system_service, $method, $node->args);
+
+      return $node;
+
+    }
+
+    return NULL;
+  }
+
+}
--- a/config/drupal-8/drupal-8.7-deprecations.yml
+++ b/config/drupal-8/drupal-8.7-deprecations.yml
@@ -2,3 +2,4 @@
 services:
   DrupalRector\Rector\Deprecation\FilePrepareDirectoryRector: ~
   DrupalRector\Rector\Deprecation\FileCreateDirectoryRector: ~
+  DrupalRector\Rector\Deprecation\FileUnmanagedCopyRector: ~
--- a/deprecation-index.yml
+++ b/deprecation-index.yml
@@ -129,3 +129,8 @@
     - DrupalRenderRootStaticUpdated.php
     - drupal_render_root.php
     - drupal_render_root_updated.php
+'file_unmanaged_copy()':
+  Rector: FileUnmanagedCopyRector.php
+  PHPStan: 'Call to deprecated function file_unmanaged_copy(). Deprecated in Drupal 8.7.0, will be removed before Drupal 9.0.0. Use \Drupal\Core\File\FileSystemInterface::copy().'
+  Examples:
+    - file_unmanaged_copy.php


rector_examples/file_unmanaged_copy.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+/**
+ * A simple example using the minimum required (all) number of arguments.
+ */
+function simple_example() {
+  $source = '/test/directory';
+  $destination = '/test/directory_2';
+  file_unmanaged_copy($source, $destination,FILE_CREATE_DIRECTORY);
```
damontgomery, line 9: @mohsin-saeed, it looks like you can actually call `file_unmanaged_copy` with just the `source` argument.

Can you create an example with that?

```
function file_unmanaged_copy($source, $destination = NULL, $replace = FILE_EXISTS_RENAME) {
  @trigger_error('file_unmanaged_copy() is deprecated in Drupal 8.7.0 and will be removed before Drupal 9.0.0. Use \Drupal\Core\File\FileSystemInterface::copy(). See https://www.drupal.org/node/3006851.', E_USER_DEPRECATED);
  try {
    $file_system = \Drupal::service('file_system');

    // Build a destination URI if necessary.
    if (!isset($destination)) {
      $destination = file_build_uri($file_system->basename($source));
    }
    return $file_system->copy($source, $destination, $replace);
  }
  catch (FileException $e) {
    return FALSE;
  }
}
```

src/Rector/Deprecation/FileUnmanagedCopyRector.php
```diff
@@ -0,0 +1,71 @@
...
+use DrupalRector\Utility\TraitsByClassHelperTrait;
+use PhpParser\Node;
+use Rector\Rector\AbstractRector;
+use Rector\RectorDefinition\CodeSample;
+use Rector\RectorDefinition\RectorDefinition;
+
+/**
+ * Replaces deprecated file_unmanaged_copy() calls.
+ *
+ * See https://api.drupal.org/api/drupal/core%21includes%21file.inc/function/file_unmanaged_copy/8.7.x.
```
damontgomery, line 14: Can you provide a link to the change record. This is usually in the function definition comment,

```
/**
 * Copies a file to a new location without database changes or hook invocation.
 *
 * This is a powerful function that in many ways performs like an advanced
 * version of copy().
 * - Checks if $source and $destination are valid and readable/writable.
 * - If file already exists in $destination either the call will error out,
 *   replace the file or rename the file based on the $replace parameter.
 * - If the $source and $destination are equal, the behavior depends on the
 *   $replace parameter. FILE_EXISTS_REPLACE will error out. FILE_EXISTS_RENAME
 *   will rename the file until the $destination is unique.
 * - Works around a PHP bug where copy() does not properly support streams if
 *   safe_mode or open_basedir are enabled.
 *   @see https://bugs.php.net/bug.php?id=60456
 *
 * @param $source
 *   A string specifying the filepath or URI of the source file.
 * @param $destination
 *   A URI containing the destination that $source should be copied to. The
 *   URI may be a bare filepath (without a scheme). If this value is omitted,
 *   Drupal's default files scheme will be used, usually "public://".
 * @param $replace
 *   Replace behavior when the destination file already exists:
 *   - FILE_EXISTS_REPLACE - Replace the existing file.
 *   - FILE_EXISTS_RENAME - Append _{incrementing number} until the filename is
 *       unique.
 *   - FILE_EXISTS_ERROR - Do nothing and return FALSE.
 *
 * @return
 *   The path to the new file, or FALSE in the event of an error.
 *
 * @deprecated in drupal:8.7.0 and is removed from drupal:9.0.0.
 *   Use \Drupal\Core\File\FileSystemInterface::copy().
 *
 * @see file_copy()
 * @see https://www.drupal.org/node/3006851
 */
```

src/Rector/Deprecation/FileUnmanagedCopyRector.php
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * What is covered:
+ * - File System Service used
+ *
+ * Improvement opportunities
+ * - Dependency Injection
+ */
+final class FileUnmanagedCopyRector extends AbstractRector {
+  use TraitsByClassHelperTrait;
```
damontgomery, line 23: Unless you are looking at traits for classes, you don't need to add this trait to the rector class.

src/Rector/Deprecation/FileUnmanagedCopyRector.php
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * See https://api.drupal.org/api/drupal/core%21includes%21file.inc/function/file_unmanaged_copy/8.7.x.
+ *
+ * What is covered:
+ * - File System Service used
+ *
+ * Improvement opportunities
+ * - Dependency Injection
+ */
+final class FileUnmanagedCopyRector extends AbstractRector {
```
damontgomery, line 22: For simple function to service method calls, we have a class you could extend, `FunctionToServiceBase`.

Unfortunately, I think that this rector rule is more complex. I'll add a comment to the overall review.

src/Rector/Deprecation/FileUnmanagedCopyRector.php
```diff
@@ -0,0 +1,71 @@
...
+    return [
+      Node\Expr\FuncCall::class,
+    ];
+  }
+
+  /**
+   * @inheritdoc
+   */
+  public function refactor(Node $node): ?Node {
+    if ($node->name instanceof Node\Name && 'file_unmanaged_copy' === (string) $node->name) {
```
damontgomery, line 55: We've started to use `if ($node->getName($node->name) === 'file_unmanaged_copy') {` in places like this which is easier to read.

damontgomery (changes requested):
Thanks for the work, @mohsin-saeed!

I think you've picked a complicated situation here since the function has an optional `$destination` argument, but the service method requires it.

We would need to account for this in our rector rule or at least list it in the "Improvement Opportunities".

```
function file_unmanaged_copy($source, $destination = NULL, $replace = FILE_EXISTS_RENAME) {
  @trigger_error('file_unmanaged_copy() is deprecated in Drupal 8.7.0 and will be removed before Drupal 9.0.0. Use \Drupal\Core\File\FileSystemInterface::copy(). See https://www.drupal.org/node/3006851.', E_USER_DEPRECATED);
  try {
    $file_system = \Drupal::service('file_system');

    // Build a destination URI if necessary.
    if (!isset($destination)) {
      $destination = file_build_uri($file_system->basename($source));
    }
    return $file_system->copy($source, $destination, $replace);
  }
  catch (FileException $e) {
    return FALSE;
  }
}
```

In the above example, we would probably need to check if `$destination` is not passed as an argument. You can check `$node->args` which is an array. If there are at least two arguments, your substitution would work. If not, we should do the more complex replacement as shown above or at least skip over that situation.

Thanks again for the submission. Do you want to make those changes?

mohsin-saeed:
> Thanks for the work, @mohsin-saeed!
> 
> I think you've picked a complicated situation here since the function has an optional `$destination` argument, but the service method requires it.
> 
> We would need to account for this in our rector rule or at least list it in the "Improvement Opportunities".
> 
> ```
> function file_unmanaged_copy($source, $destination = NULL, $replace = FILE_EXISTS_RENAME) {
>   @trigger_error('file_unmanaged_copy() is deprecated in Drupal 8.7.0 and will be removed before Drupal 9.0.0. Use \Drupal\Core\File\FileSystemInterface::copy(). See https://www.drupal.org/node/3006851.', E_USER_DEPRECATED);
>   try {
>     $file_system = \Drupal::service('file_system');
> 
>     // Build a destination URI if necessary.
>     if (!isset($destination)) {
>       $destination = file_build_uri($file_system->basename($source));
>     }
>     return $file_system->copy($source, $destination, $replace);
>   }
>   catch (FileException $e) {
>     return FALSE;
>   }
> }
> ```
> 
> In the above example, we would probably need to check if `$destination` is not passed as an argument. You can check `$node->args` which is an array. If there are at least two arguments, your substitution would work. If not, we should do the more complex replacement as shown above or at least skip over that situation.
> 
> Thanks again for the submission. Do you want to make those changes?

Thank you for review @damontgomery. I will make these changes along with your other mentioned changes.

src/Rector/Deprecation/FileUnmanagedCopyRector.php
```diff
@@ -0,0 +1,71 @@
+<?php
+
+namespace DrupalRector\Rector\Deprecation;
+
+use DrupalRector\Utility\TraitsByClassHelperTrait;
+use PhpParser\Node;
+use Rector\Rector\AbstractRector;
+use Rector\RectorDefinition\CodeSample;
+use Rector\RectorDefinition\RectorDefinition;
```
damontgomery, line 9: @mohsin-saeed, with the update to Rector, all of the `Rector\` class names need to be changed to `Rector\Core\`.

Sorry for moving things from underneath you. :(

PR #47, opened 2020-04-19
Path alias service name rectors.

damontgomery:


damontgomery:
@dbuytaert,

I made a branch with your code here and then applied some suggested updates.

I'll comment on the PR with some explanations.

Thanks for taking a look and for submitting your rector rule!

config/drupal-8/drupal-8.8-deprecations.yml
```diff
@@ -0,0 +1,6 @@
+# Contains automatic fixes for some deprecations introduced in Drupal 8.8.
+services:
+  DrupalRector\Rector\Deprecation\PathAliasManagerServiceNameRector: ~
+  DrupalRector\Rector\Deprecation\PathAliasWhitelistServiceNameRector: ~
+  DrupalRector\Rector\Deprecation\PathSubscriberServiceNameRector: ~
+  DrupalRector\Rector\Deprecation\PathProcessorAliasServiceNameRector: ~
```
damontgomery, line 6: We are creating one configuration line per rector rule and each rule changes one thing if possible. There are other ways to go about this, but we thought this would be a simple way to start.

rector_examples/path_alias_manager.php
```diff
@@ -0,0 +1,12 @@
...
+/**
+ * This demonstrates the deprecated static calls that might be called from procedural code like `.module` files.
+ */
+
+/**
+ * A simple example using the minimum number of arguments.
+ */
+function simple_example() {
+  \Drupal::service('path.alias_manager');
+}
```
damontgomery, line 12: We would like to have an example file for each deprecation.

rector_examples/path_alias_manager_updated.php
```diff
@@ -0,0 +1,12 @@
...
+/**
+ * This demonstrates the deprecated static calls that might be called from procedural code like `.module` files.
+ */
+
+/**
+ * A simple example using the minimum number of arguments.
+ */
+function simple_example() {
+  \Drupal::service('path_alias.manager');
+}
```
damontgomery, line 12: We would like to have another example file that shows how the code would be updated.

src/Rector/Deprecation/DrupalServiceRenameBase.php
```diff
@@ -0,0 +1,48 @@
...
+use Rector\Core\RectorDefinition\CodeSample;
+use Rector\Core\RectorDefinition\RectorDefinition;
+
+/**
+ * Renames deprecated Drupal::services().
+ *
+ * What is covered:
+ *  - Renames argument in Drupal::service() calls.
+ */
+abstract class DrupalServiceRenameBase extends StaticArgumentRenameBase {
```
damontgomery, line 15: Since this seems like a common use case that you pointed out, I created this base class. This base class is built off the `StaticArgumentRenameBase` which could have broader use for other methods and classes. The two have the split pieces of the code you sent.

src/Rector/Deprecation/PathAliasManagerServiceNameRector.php
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Renames deprecated Drupal::services('path.alias_manager') argument.
+ *
+ * See https://www.drupal.org/node/3092086 for change record.
+ *
+ * What is covered:
+ *  - Replacement in Drupal::service() calls.
+ */
+final class PathAliasManagerServiceNameRector extends DrupalServiceRenameBase {
```
damontgomery, line 18: For now (we may revisit this), we've tried to keep one rector class per deprecation. This seems the simplest to jump into even if it's a bit verbose in cases. We may revisit this later.

src/Rector/Deprecation/StaticArgumentRenameBase.php
```diff
@@ -0,0 +1,76 @@
...
+
+use PhpParser\Node;
+use Rector\Core\Rector\AbstractRector;
+use Rector\Core\RectorDefinition\CodeSample;
+use Rector\Core\RectorDefinition\RectorDefinition;
+
+/**
+ * Renames deprecated the argument in a static call.
+ */
+abstract class StaticArgumentRenameBase extends AbstractRector {
```
damontgomery, line 13: This is the general use base class.

src/Rector/Deprecation/StaticArgumentRenameBase.php
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  /**
+   * @inheritdoc
+   */
+  public function refactor(Node $node): ?Node {
+
+    if ($node instanceof Node\Expr\StaticCall) {
+      /** @var Node\Expr\StaticCall $node */
+      if ($this->getName($node->name) === $this->methodName && (string) $node->class === $this->fullyQualifiedClassName) {
```
damontgomery, line 59: `$this->getName($node->name) === $this->methodName` is the way we've started checking the names of nodes. It seems a little more straight forward than the other ways we've used.

src/Rector/Deprecation/StaticArgumentRenameBase.php
```diff
@@ -0,0 +1,76 @@
...
+
+        if (count($node->args) === 1) {
+          /* @var Node\Arg $argument */
+          $argument = $node->args[0];
+
+          if ($argument->value instanceof Node\Scalar\String_ && $argument->value->value === $this->deprecatedArgument) {
+            $node->args[0] = new Node\Arg(new Node\Scalar\String_($this->argument));
```
damontgomery, line 66: Since we are only changing the parameters, we can modify the node directly instead of creating a new one.

src/Rector/Deprecation/StaticArgumentRenameBase.php
```diff
@@ -0,0 +1,76 @@
...
+
+            return $node;
```
damontgomery, line 68: Returning the node makes sure we get output like,

```
Applied rules:

 * DrupalRector\Rector\Deprecation\PathProcessorAliasServiceNameRector
```

It will still make the update without this, but it seems like a good practice.

damontgomery (commented):
@dbuytaert,

Here are the comments about the updates I made.

config/drupal-8/drupal-8.8-deprecations.yml
dbuytaert: I started with one Rector per rule but found it to be repetitive: it takes more time and effort to write, and I didn't immediately see the advantage.  I was optimizing for speed and simplicity, rather than abstraction.  One Rector per rule certainly works, and might show its advantages in time. Not against it at all, and consistent with the code that already exists. 👍

src/Rector/Deprecation/DrupalServiceRenameBase.php
dbuytaert: The Base classes are great.  It took me a while to realize they existed; they get lost in between the non-base classes.  I didn't realize until now that class names ending with *Base.php were all the  re-usable utility classes and that I could find them by looking at the class names.  D'oh!  One developer experience (DX) improvement could be to put them in their own sub-package.

src/Rector/Deprecation/StaticArgumentRenameBase.php
dbuytaert: The Base classes are great.  It took me a while to realize they existed; they get lost in between the non-base classes.  I didn't realize until now that class names ending with *Base.php were all the  re-usable utility classes and that I could find them by looking at the class names.  D'oh!  One developer experience (DX) improvement could be to put them in their own sub-package.

src/Rector/Deprecation/StaticArgumentRenameBase.php
dbuytaert: 👍 I was going to try that, but didn't get around doing that yet. Glad to know that works!

dbuytaert (approved):
All changes look good to me!

config/drupal-8/drupal-8.8-deprecations.yml
damontgomery: Thanks for the feedback!

src/Rector/Deprecation/DrupalServiceRenameBase.php
damontgomery: Thanks for the feedback!

I was finding that a bit as well. We're getting a lot more files now. I'll discuss this with our internal team.

damontgomery:
Thanks, @dbuytaert.

Closing for your PR #48 since it extends what we have here and you can modify that one.

PR #50, opened 2020-04-21
Test current branch (also for forked repo)

shaal:


shaal:
@damontgomery 
This is finally allowing testing on branches and forks, it finds errors before a PR is being merged.

The key is using local composer packages, more details here: https://medium.com/pvtl/local-composer-package-development-47ac5c0e5bb4

example of (intentionally) failed test in the fork:
https://github.com/shaal/drupal-rector/runs/604435383?check_suite_focus=true#step:10:8

example of (intentionally) failed test on the PR on main repo:
https://github.com/palantirnet/drupal-rector/runs/604435387?check_suite_focus=true#step:10:8


P.s
We should probably modify `drupal-rector-sandbox` to take advantage of local composer packages as well.

damontgomery:
I think this makes sense, @shaal. Would it make sense to update this PR and have it replace the current Github Workflow?

shaal:
@damontgomery I made the changes we have discussed, it's ready for review.

.github/workflows/drupal_rector_examples_as_local_package.yml
```diff
@@ -0,0 +1,46 @@
+name: drupal_rector_examples_as_local_package
+
+# This test will run on every pull request, and on every commit on any branch
```
damontgomery, line 3: Thanks

.github/workflows/drupal_rector_examples_as_packagist_branch.yml
```diff
@@ -1,13 +1,14 @@
-name: drupal_rector_examples
+name: drupal_rector_examples_as_packagist_branch
 
+# This test will run on every pull request, and on every commit that is pushed to master
```
damontgomery, line 3: Thanks.

damontgomery (approved):
Thanks, @shaal , this makes sense. The comments are helpful and the names are clear.

PR #62, opened 2020-05-01
Rector to replace const REQUEST_TIME with a Drupal time service.

tobiasbaehr:


shaal:
I added a comment here - https://www.drupal.org/project/rector/issues/3121405#comment-13592893

I think that we can add this rule now without waiting for that improvement.

damontgomery (approved):
Thanks for this work @tobiasbaehr!

I'm going to let the automated tests finish and merge assuming all things work out.

PR #63, opened 2020-05-01
render_root of renderer service has only one argument.

tobiasbaehr:
See https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Render%21Renderer.php/function/Renderer%3A%3ArenderRoot/8.9.x


damontgomery:
@tobiasbaehr, thanks for catching this!

PR #64, opened 2020-05-01
A rector to replace deprecated file_default_scheme function calls.

tobiasbaehr:
For this use case I created an abstract class FunctionToImmutableConfigBase to reuse it for other use cases.

### DIFF
--- a/src/Rector/Deprecation/Base/FunctionToImmutableConfigBase.php
+++ b/src/Rector/Deprecation/Base/FunctionToImmutableConfigBase.php
@@ -1,0 +1,47 @@
+<?php declare(strict_types=1);
+
+namespace DrupalRector\Rector\Deprecation\Base;
+
+use PhpParser\Node;
+use Rector\Core\Rector\AbstractRector;
+
+
+/**
+ * Replace deprecated function calls with a wrapped config.factory service call to get an immutable configuration.
+ *
+ * @see \DrupalRector\Rector\Deprecation\FileDefaultScheme for an example.
+ */
+abstract class FunctionToImmutableConfigBase extends AbstractRector
+{
+    /**
+     * @inheritDoc
+     */
+    public function getNodeTypes(): array
+    {
+        return [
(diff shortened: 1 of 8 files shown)


src/Rector/Deprecation/Base/FunctionToImmutableConfigBase.php
```diff
@@ -0,0 +1,47 @@
...
+
+namespace DrupalRector\Rector\Deprecation\Base;
+
+use PhpParser\Node;
+use Rector\Core\Rector\AbstractRector;
+
+
+/**
+ * Replace deprecated function calls with a wrapped config.factory service call to get an immutable configuration.
+ *
```
damontgomery, line 11: Can you add a comment describing the scope of this base class like the example below?

```
What is covered:
 - Static replacement
Improvement opportunities
 - Dependency injection
```

src/Rector/Deprecation/FileDefaultScheme.php
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Replaces the deprecated file_default_scheme function calls.
+ *
+ * @see https://www.drupal.org/node/3049030 for change record.
+ */
+final class FileDefaultScheme extends FunctionToImmutableConfigBase
+{
+    protected $deprecatedFunctionName = 'file_default_scheme';
+    protected $configObject = 'system.file';
+    protected $configName = 'default_scheme';
```
damontgomery, line 20: These variables make sense with the names. Can you also define these in the base class to help people understand which ones are expected? You can look at `FunctionToServiceBase` for an example.

damontgomery (changes requested):
@tobiasbaehr, thanks so much. I think the base class is a great idea here and it makes sense to me.

I added two small comments about clarifying the code with comments and base class properties. Can you take a look?

damontgomery:
Thanks @tobiasbaehr! I merged this in. I made a small adjustment to the index file. Looks good.

PR #65, opened 2020-05-04
Fix errors in error index

goba:
After I started tracking whether deprecation_status finds the errors from the index, I identified these 3 broken. The first two have an evident mistake at the end, the third one has breaking spaces around the backslashes.


shaal (approved):
Thank you @goba !

Issue #66, opened 2020-05-05
Got iconv errors in Alpine based Docker image


back-2-95:
And got rid off the error with help of
https://github.com/nunomaduro/phpinsights/issues/43

So this could be put to some FAQ section or something like that.

shaal:
I am not familiar with Apline based Docker images.
Can you create a PR with the text that we should add to FAQ?

back-2-95:
@shaal I created PR #67

PR #67, opened 2020-05-06
Docs about Iconv error when running Rector in Alpine Docker

back-2-95:
Closes issue #66


shaal:
Thank you @back-2-95 

@damontgomery would it be ok to add this to README?

damontgomery:
I think this would be helpful for now.

If we get a ton of troubleshooting suggestions we can always split the file off.

Thanks.

PR #73, opened 2020-05-10
KernelTestBase getmock error incorrect

goba:


goba:
Also BrowserTestBase getmock does not appear as an error in https://dev.acquia.com/drupal9/deprecation_status/errors?message=BrowserTestBase&project=&category=&group=&op=Submit however it seems to have coverage in here, but the error message listed for that is also wrong. Is that an actual method / error that can happen?

goba:
Fixed two more wrong messages. A bit of a scope creep sorry :D

damontgomery:
Thanks for fixing these @goba!

goba:
What do you think of the BrowserTestBase getmock question? Because I am not finding an error for it but there is a rector for it? https://github.com/palantirnet/drupal-rector/pull/73#issuecomment-626298690

damontgomery:
@goba, according to https://www.drupal.org/node/2907725 it is a possible issue.

goba:
Ok proposed pull request in https://github.com/palantirnet/drupal-rector/pull/78

PR #74, opened 2020-05-11
entity_load, node_load, file_load, user_load rector rules.

damontgomery:


damontgomery:
I see there is a node_load and file_load that are also deprecated. I'm going to try and make this a base class and work on those as well.

PR #78, opened 2020-05-12
BrowserTestBase getmock incorrect in deprecation index

goba:
As discussed earlier in https://github.com/palantirnet/drupal-rector/pull/73


damontgomery:
Thanks, @goba!

PR #79, opened 2020-05-12
Add file_uri_target() rector.

miromichalicka:
Adding rector for file_uri_target().


damontgomery:
Thanks so much @miromichalicka! Looks great.

PR #81, opened 2020-05-13
Add rector for method l() of class ControllerBase.

Kbentham:
https://palantir.atlassian.net/browse/REC-89

### DIFF
--- a/rector_examples_updated/src/ControllerBaseLWithTrait.php
+++ b/rector_examples_updated/src/ControllerBaseLWithTrait.php
@@ -1,0 +1,35 @@
+<?php
+
+namespace Drupal\rector_examples;
+
+use Drupal\Core\Link;
+use Drupal\Core\Url;
+use Drupal\Core\Routing\LinkGeneratorTrait;
+
+/**
+ * Example of static calls from a class with the trait.
+ */
+class ControllerBaseLWithTrait {
+
+  use LinkGeneratorTrait;
+
+  /**
+   * A simple example using the minimum number of arguments.
+   */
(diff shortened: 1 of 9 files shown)


src/Rector/Deprecation/ControllerBaseLRector.php
```diff
@@ -0,0 +1,79 @@
...
+ * See https://www.drupal.org/node/2614344 for change record.
+ *
+ * What is covered:
+ * - Trait usage when the `LinkGeneratorTrait` is already present on the class
+ *
+ * Improvement opportunities
+ * - Add trait for classes
+ *   - `use LinkGeneratorTrait;`
+ */
+final class ControllerBaseLRector extends AbstractRector
```
damontgomery: @Kbentham, can we name this (and the examples) `LinkGeneratorTraitL`, etc? I think this started as only dealing with the ControllerBase class, but now includes more.

rector_examples_updated/src/ControllerBaseLWithTrait.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+namespace Drupal\rector_examples;
+
+use Drupal\Core\Link;
+use Drupal\Core\Url;
```
damontgomery: The Behat tests are failing because the `use` statements are in a different order. Can you see if the local version produces the same thing as the CI tool. If not, maybe we hand switch this to match the CI tool. :(

damontgomery:
Kelsey, I left a few comments, but this looks like a great approach overall. Thanks!

PR #83, opened 2020-05-14
Fix deprecation-index.yml invisible unicode and non-escaped quotes

tedbow:
Somehow some invisible unicode character got in `deprecation-index.yml`

I found this when I was trying to compare strings produced by phpstan and the strings in this yml file.

### DIFF
--- a/deprecation-index.yml
+++ b/deprecation-index.yml
@@ -62,7 +62,7 @@
   Rector: BrowserTestBaseGetMock.php
   PHPStan: 'Call to deprecated method getMock() of class Drupal\Tests\BrowserTestBase. Deprecated in drupal:8.5.0 and is removed from drupal:9.0.0. Use Drupal\Tests\PhpunitCompatibilityTrait::createMock() instead.'
   Examples:
-      - BrowserTestBaseGetMock.php
+    - BrowserTestBaseGetMock.php
 'KernelTestBase::getMock()':
   Rector: KernelTestBaseGetMock.php
   PHPStan: 'Call to deprecated method getMock() of class Drupal\KernelTests\KernelTestBase. Deprecated in drupal:8.5.0 and is removed from drupal:9.0.0. Use Drupal\Tests\PhpunitCompatibilityTrait::createMock() instead.'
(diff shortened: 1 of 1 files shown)


deprecation-index.yml
```diff
@@ -54,7 +54,7 @@
     - FilePrepareDirectoryStatic.php
 'file_scan_directory()':
   Rector: FileScanDirectoryRector.php
-  PHPStan: 'Call to deprecated function file_scan_directory(). Deprecated in drupal:8.8.0 and is removed from drupal:9.0.0. Use Drupal​\​Core​\​File​\​FileSystemInterface::scanDirectory() instead.'
+  PHPStan: 'Call to deprecated function file_scan_directory(). Deprecated in drupal:8.8.0 and is removed from drupal:9.0.0. Use Drupal\Core\File\FileSystemInterface::scanDirectory() instead.'
```
tedbow, line 57: You can see there is difference in the slash in the line but there is no visual difference. This appears to be the character https://www.fileformat.info/info/unicode/char/200b/index.htm

The only way I could see it was editing the file in vim🤷‍♂️.
Also doing a `git diff` on the command line after editing shows the unicode character number.
Screenshot 2020-05-13 23 11 45

tedbow: I searched and I didn't see anymore of these characters in the project.

deprecation-index.yml
```diff
@@ -162,18 +162,18 @@
     - EntityGetFormDisplayStatic.php
 'file_default_scheme()':
   Rector: FileDefaultSchemeRector.php
-  PHPStan: 'Call to deprecated function file_default_scheme(). Deprecated in drupal:8.8.0 and is removed from drupal:9.0.0. Use Drupal::config('system.file')->get('default_scheme') instead.'
+  PHPStan: "Call to deprecated function file_default_scheme(). Deprecated in drupal:8.8.0 and is removed from drupal:9.0.0. Use Drupal::config('system.file')->get('default_scheme') instead."
```
tedbow: There was single quote inside here so changed to using double quote so we don't have escape

This keeps the file from actually being able to be parsed as yml.

damontgomery: Thanks, @tedbow.

I'm going to try and write a quick Behat test for this so it doesn't happen again. This isn't the first time. :(

Can I update your PR directly to use `''` for escaping? I found that to be more affective since so many of these have `\` and other characters that makes `"` quotes kind of a mess.

tedbow: Sure yeah update it however you see fit.

damontgomery:
Thanks, @tedbow!

Hoping that the automated tests will help us moving forward.

PR #84, opened 2020-05-14
Test for valid deprecation-index.yml.

damontgomery:
Check that we have a valid YAML file.
Check that there are no non-ASCII characters.


damontgomery:
This is going to fail, because our index file isn't valid yet.

PR #89, opened 2020-05-21
Add comments to updated files.

damontgomery:
This is intended to highlight cases where Rector covers the most common use cases, but may have made a substitution that was incorrect in edge cases. Or, when things are weird.


damontgomery:
We're collecting feedback on the general approach. I can probably update this next week and address the merge conflicts.

Issue #94, opened 2020-05-22
Error: Pick more specific node than "PhpParser\Node\Expr\MethodCall"


shaal:
Steps to replicate:
1. `git clone git@github.com:palantirnet/drupal-rector-sandbox.git`
1. `composer require drupal/contentserialize`
1. `vendor/bin/rector process --dry-run web/modules/contrib/contentserialize/tests/src/Unit/OptionsProviderTest.php`

Throw this error:
```
[ERROR] Could not process "web/modules/contrib/contentserialize/tests/src/Unit/OptionsProviderTest.php" file, due to:
         "Pick more specific node than "PhpParser\Node\Expr\MethodCall"".
```

@Bysa Boro found that changing in `web/modules/contrib/contentserialize/tests/src/Unit/OptionsProviderTest.php`

This code:
```
$this->config = $this->getMockBuilder(ImmutableConfig::class)
      ->disableOriginalConstructor()
      ->getMock();
```

To:
```
$mockBuilder = $this->getMockBuilder(ImmutableConfig::class);
    $mockBuilder->disableOriginalConstructor();
    $this->config = $mockBuilder->getMock();
```

Results in Rector running successfully without that error message.

damontgomery:
Thanks for the detailed bug report. The example code was helpful. I've added this to our set of examples so we should catch similar issues going forward.

See #98.

PR #98, opened 2020-05-26
Fix issue with GetMockBase.

damontgomery:
When this was called on a method, like in the new example, an error was thrown.

The new conditional logic checks that we are acting on the `this` which is a variable.


damontgomery:
This should fix #94 .

PR #104, opened 2020-06-02
Issue #3144659 by dww: Add rector rule for DATETIME_STORAGE_FORMAT

dww:
See https://www.drupal.org/project/rector/issues/3144659

Thanks!
-Derek

### DIFF
--- a/rector_examples/datetime_datetime_storage_format.php
+++ b/rector_examples/datetime_datetime_storage_format.php
@@ -1,0 +1,21 @@
+<?php
+
+/**
+ * This demonstrates the deprecated static calls that might be called from procedural code like `.module` files.
+ */
+
+/**
+ * A simple example.
+ */
+function simple_example() {
+  $format = DATETIME_DATETIME_STORAGE_FORMAT;
+
+}
+
+/**
+ * An example using the constant as an argument.
+ */
+function as_an_argument() {
+  $date = new DrupalDateTime('now', new \DateTimezone('America/Los_Angeles'));
+  $now = $date->format(DATETIME_DATETIME_STORAGE_FORMAT);
+}
--- a/rector_examples/datetime_datetime_storage_format.php
+++ b/rector_examples/datetime_datetime_storage_format.php
@@ -1,15 +1,17 @@
 <?php
 
 /**
- * This demonstrates the deprecated static calls that might be called from procedural code like `.module` files.
+ * @file
+ * Demonstrates deprecated constants that might be used in procedural code.
  */
+
+use Drupal\Core\Datetime\DrupalDateTime;
 
 /**
  * A simple example.
  */
 function simple_example() {
   $format = DATETIME_DATETIME_STORAGE_FORMAT;
-
 }
 
 /**
--- a/rector_examples_updated/datetime_datetime_storage_format.php
+++ b/rector_examples_updated/datetime_datetime_storage_format.php
@@ -1,0 +1,22 @@
+<?php
+
+use Drupal\datetime\Plugin\Field\FieldType\DateTimeItemInterface;
+/**
+ * This demonstrates the deprecated static calls that might be called from procedural code like `.module` files.
+ */
+
+/**
+ * A simple example.
+ */
+function simple_example() {
+  $format = DateTimeItemInterface::DATETIME_STORAGE_FORMAT;
+
+}
+
+/**
+ * An example using the constant as an argument.
+ */
+function as_an_argument() {
+  $date = new DrupalDateTime('now', new \DateTimezone('America/Los_Angeles'));
+  $now = $date->format(DateTimeItemInterface::DATETIME_STORAGE_FORMAT);
+}
--- a/rector_examples_updated/datetime_datetime_storage_format.php
+++ b/rector_examples_updated/datetime_datetime_storage_format.php
@@ -1,12 +1,12 @@
 <?php
 
+use Drupal\datetime\Plugin\Field\FieldType\DateTimeItemInterface;
 /**
  * @file
- * Demonstrates updated code using the new constants.
+ * Demonstrates deprecated constants that might be used in procedural code.
  */
 
 use Drupal\Core\Datetime\DrupalDateTime;
-use Drupal\datetime\Plugin\Field\FieldType\DateTimeItemInterface;
 
 /**
  * A simple example.
--- a/rector_examples_updated/datetime_datetime_storage_format.php
+++ b/rector_examples_updated/datetime_datetime_storage_format.php
@@ -1,16 +1,18 @@
 <?php
 
-use Drupal\datetime\Plugin\Field\FieldType\DateTimeItemInterface;
 /**
- * This demonstrates the deprecated static calls that might be called from procedural code like `.module` files.
+ * @file
+ * Demonstrates updated code using the new constants.
  */
+
+use Drupal\Core\Datetime\DrupalDateTime;
+use Drupal\datetime\Plugin\Field\FieldType\DateTimeItemInterface;
 
 /**
  * A simple example.
  */
 function simple_example() {
   $format = DateTimeItemInterface::DATETIME_STORAGE_FORMAT;
-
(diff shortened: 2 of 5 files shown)


rector_examples/datetime_datetime_storage_format.php
```diff
@@ -0,0 +1,21 @@
...
+function simple_example() {
+  $format = DATETIME_DATETIME_STORAGE_FORMAT;
+
+}
+
+/**
+ * An example using the constant as an argument.
+ */
+function as_an_argument() {
+  $date = new DrupalDateTime('now', new \DateTimezone('America/Los_Angeles'));
```
dww, line 21: Do we need this line?

rector_examples_updated/datetime_datetime_storage_format.php
```diff
@@ -0,0 +1,22 @@
...
+function simple_example() {
+  $format = DateTimeItemInterface::DATETIME_STORAGE_FORMAT;
+
+}
+
+/**
+ * An example using the constant as an argument.
+ */
+function as_an_argument() {
+  $date = new DrupalDateTime('now', new \DateTimezone('America/Los_Angeles'));
```
dww, line 22: And here.

src/Rector/Deprecation/DatetimeDatetimeStorageFormatRector.php
```diff
@@ -0,0 +1,48 @@
...
+
+  /**
+   * @inheritdoc
+   */
+  public function getDefinition(): RectorDefinition
+  {
+    return new RectorDefinition('Fixes deprecated DATETIME_DATETIME_STORAGE_FORMAT use',[
+      new CodeSample(
+        <<<'CODE_BEFORE'
+$date = new DrupalDateTime('now', new \DateTimezone('America/Los_Angeles'));
```
dww, line 38: Here.

src/Rector/Deprecation/DatetimeDatetimeStorageFormatRector.php
```diff
@@ -0,0 +1,48 @@
...
+$now = $date->format(DATETIME_DATETIME_STORAGE_FORMAT);
+CODE_BEFORE
+        ,
+        <<<'CODE_AFTER'
+$date = new DrupalDateTime('now', new \DateTimezone('America/Los_Angeles'));
```
dww, line 45: Here

dww (commented):
Does the example code need to be valid or can we leave these out? Thanks! -Derek

rector_examples/datetime_datetime_storage_format.php
damontgomery: @dww, Since this is a deprecation for a constant, assigning it to a variable seems fine as a minimum test.

If you want to illustrate a real use case, that's always great. We don't have hard rules, but we would like the examples to provide some context if it's helpful.

Do we need a `use` statement or a fully qualified namespace `Drupal\Core\Datetime\DrupalDateTime` so this is valid code?

Rector does try to resolve classes.

src/Rector/Deprecation/DatetimeDatetimeStorageFormatRector.php
damontgomery: This is used only for generating documentation as far as I know, but it would be good if it was valid. Do we need to use the fully qualified class name?

damontgomery (changes requested):
@dww, Thanks so much for this.

I've left a few comments.

If you want to keep the extra example, I think it's close. Can you confirm if a fully qualified class name or use statement is needed?

If you don't want the extra example, the simple one is good as well.

Thanks again.

dww:
> Do we need a use statement or a fully qualified namespace Drupal\Core\Datetime\DrupalDateTime so this is valid code?

Either is valid.  Usually we do a use in core code.

Pushed cd8af284dc7cd96a22e08025f3b351dd3ac387f7 to address it.  That's the only spot we used FQCN. Everything else already has a use.

Why is "use Drupal\datetime\Plugin\Field\FieldType\DateTimeItemInterface;" getting added to my files directly below the ?php tag, before @file comment, and no where near sorted in the existing use statements (if any).  Is that just how rector works or am I doing something wrong?

Thanks!

src/Rector/Deprecation/DatetimeDatetimeStorageFormatRector.php
dww: Fixed in cd8af284dc7cd96a22e08025f3b351dd3ac387f7

rector_examples/datetime_datetime_storage_format.php
dww: Yup, it already has the use.

damontgomery:
> > Do we need a use statement or a fully qualified namespace Drupal\Core\Datetime\DrupalDateTime so this is valid code?
> 
> Either is valid. Usually we do a use in core code.
> 
> Pushed [cd8af28](https://github.com/palantirnet/drupal-rector/commit/cd8af284dc7cd96a22e08025f3b351dd3ac387f7) to address it. That's the only spot we used FQCN. Everything else already has a use.
> 
> Why is "use Drupal\datetime\Plugin\Field\FieldType\DateTimeItemInterface;" getting added to my files directly below the ?php tag, before @file comment, and no where near sorted in the existing use statements (if any). Is that just how rector works or am I doing something wrong?
> 
> Thanks!

Thanks!

There is a set of configuration options you can enable or disable in `rector.yml` to add `use` statements.

```
# Create `use` statements.
auto_import_names: true
# Do not convert `\Drupal` to `Drupal`, etc.
import_short_classes: false
```

This is part of Rector and I'm not sure exactly how it works. It seems to mostly do what we want, but does have some weird behavior like what you mentioned. If you prefer to use FQCN, you can turn that off.

For Drupal Rector, we have it turned on and all the Rectors pass FQCN in the code. I wasn't able to figure out another approach beyond the all-or-nothing option here.

rector_examples/datetime_datetime_storage_format.php
damontgomery, line 21: Can you use a `use` statement or FQCN here for `DrupalDateTime`?

src/Rector/Deprecation/DatetimeDatetimeStorageFormatRector.php
```diff
@@ -0,0 +1,49 @@
...
+
+  /**
+   * @inheritdoc
+   */
+  public function getDefinition(): RectorDefinition
+  {
+    return new RectorDefinition('Fixes deprecated DATETIME_DATETIME_STORAGE_FORMAT use',[
+      new CodeSample(
+        <<<'CODE_BEFORE'
+$date = new DrupalDateTime('now', new \DateTimezone('America/Los_Angeles'));
```
damontgomery, line 38: Can you use a `use` statement or FQCN here for `DrupalDateTime`?

damontgomery (changes requested):
Thanks, @dww.

I think it's valuable to have examples here that are functional. I couldn't find the class names in use statements and would like those to be there. Can you take a look?

rector_examples/datetime_datetime_storage_format.php
dww: Sorry, I misunderstood. Thought you were talking about the new interfacve. Didn't notice DrupalDateTime. Fixed in 734a30c.

src/Rector/Deprecation/DatetimeDatetimeStorageFormatRector.php
dww: Fixed in 734a30c, thanks!

dww:
Not sure how to tell GH I've addressed all the requested changes. ;)  I think this is ready. 

Please let me know if there's anything else.

Thanks!
-Derek

dww:
Thanks, tests. ;)  I fixed the trouble in commit 9265ee2
Now I think this is ready...

damontgomery (approved):
Thanks again, @dww!

PR #105, opened 2020-06-03
3144659 datetime storage timezone

dww:
Should apply cleaning on top of the other PR.

### DIFF
--- a/rector_examples/datetime_storage_timezone.php
+++ b/rector_examples/datetime_storage_timezone.php
@@ -1,0 +1,19 @@
+<?php
+
+/**
+ * This demonstrates the deprecated static calls that might be called from procedural code like `.module` files.
+ */
+
+/**
+ * A simple example.
+ */
+function simple_example() {
+  $timezone = DATETIME_STORAGE_TIMEZONE;
+}
+
+/**
+ * An example using the constant as an argument.
+ */
+function as_an_argument() {
+  $now->setTimezone(new \DateTimeZone(DATETIME_STORAGE_TIMEZONE));
+}
--- a/src/Rector/Deprecation/DatetimeStorageTimezoneRector.php
+++ b/src/Rector/Deprecation/DatetimeStorageTimezoneRector.php
@@ -1,0 +1,47 @@
+<?php
+
+namespace DrupalRector\Rector\Deprecation;
+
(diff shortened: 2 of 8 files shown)


dww:
Or is it better to cleanly put this commit in the other CR and close this?

rector_examples/datetime_storage_timezone.php
```diff
@@ -0,0 +1,19 @@
...
+ */
+function simple_example() {
+  $timezone = DATETIME_STORAGE_TIMEZONE;
+}
+
+/**
+ * An example using the constant as an argument.
+ */
+function as_an_argument() {
+  $now->setTimezone(new \DateTimeZone(DATETIME_STORAGE_TIMEZONE));
```
damontgomery, line 18: Can you define $now so that this is valid code?

src/Rector/Deprecation/DatetimeStorageTimezoneRector.php
```diff
@@ -0,0 +1,47 @@
...
+
+  /**
+   * @inheritdoc
+   */
+  public function getDefinition(): RectorDefinition
+  {
+    return new RectorDefinition('Fixes deprecated DATETIME_STORAGE_TIMEZONE use',[
+      new CodeSample(
+        <<<'CODE_BEFORE'
+$now->setTimezone(new \DateTimeZone(DATETIME_STORAGE_TIMEZONE));
```
damontgomery, line 37: Can you define $now in this example or use the simple `$timezone = DATETIME_STORAGE_TIMEZONE` example? Same for the after.

damontgomery (changes requested):
Thanks again, @dww.

Either one PR or multiple is OK with me. If it's OK with you and there are merge conflicts between multiple PRs, I can try to push changes to your fork to resolve them before merging in the code.

For this PR (similar to the other), I want to avoid confusion, particularly with examples that aren't functional. I'm not sure that it would cause issues with Rector, but I think people might be confused by the example if it includes references to variables that were never created, etc. I'd rather we only include simple examples than add additional ones that are confusing.

Can you check the comments and let me know if you have suggestions / questions?

Thanks again.

dww:
Sorry, I screwed up my forked repo a bit.  I'll open a new PR with clean commits now that the other PR is merged.  I'll fix these comments there, too.  Thanks!

PR #112, opened 2020-06-21
EntityType::GetSingularLabel Rector.

CocoaBean5:
- Created a new rector rule for https://www.drupal.org/node/3075567
Screen Shot 2020-06-21 at 5 09 34 PM

## Note
- Method chaining results in an error when running `vendor/bin/rector process web/modules/custom/rector_examples --dry-run`. This comment has been added to the example code.

### DIFF
--- a/deprecation-index.yml
+++ b/deprecation-index.yml
@@ -253,3 +253,8 @@
   PHPStan: 'Call to deprecated constant DATETIME_STORAGE_TIMEZONE: Deprecated in drupal:8.5.0 and is removed from drupal:9.0.0. Use Drupal\datetime\Plugin\Field\FieldType\DateTimeItemInterface::STORAGE_TIMEZONE instead.'
   Examples:
     - datetime_storage_timezone.php
+'EntityType::getLowercaseLabel()':
+  Rector: EntityTypeGetSingularLabelRector.php
+  PHPStan: 'Call to deprecated method EntityType::getLowercaseLabel(). Deprecated in drupal:8.8.0 and is removed from drupal:9.0.0. Use EntityType::getSingularLabel().'
+  Examples:
+    - entity_type_getsingularlabel.php
--- a/rector_examples/entity_type_getsingularlabel.php
+++ b/rector_examples/entity_type_getsingularlabel.php
@@ -1,0 +1,26 @@
+<?php
+
+/**
+ * EntityType::getSingularLabel() now handles lower-casing the entity type label; getLowercaseLabel() deprecated.
+ */
+
+/**
+ * A simple example.
+ */
+function simple_example() {
+  /* @var \Drupal\node\Entity\Node $node */
+  $node = \Drupal::entityTypeManager()->getStorage('node')->load(123);
+  $entity_type = $node->getEntityType();
+  $entity_type->getLowercaseLabel();
+}
+
+/**
+ * A chained example.
+ *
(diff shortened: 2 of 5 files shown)


rector_examples/entity_type_getsingularlabel.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+/**
+ * EntityType::getSingularLabel() now handles lower-casing the entity type label; getLowercaseLabel() deprecated.
+ */
```
damontgomery, line 5: Thanks for the comment!

rector_examples/entity_type_getsingularlabel.php
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * A chained example.
+ *
+ * Seems like we should support this, but the method chaining errors out.
+ */
+//function chained_example() {
+//  /* @var \Drupal\node\Entity\Node $node */
+//  $node = \Drupal::entityTypeManager()->getStorage('node')->load(123);
+//  $label = $node->getEntityType()->getLowercaseLabel();
```
damontgomery, line 25: Thanks, I think if you can leave the example in here, we can work on the rule!

rector_examples_updated/entity_type_getsingularlabel.php
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * A simple example.
+ */
+function simple_example() {
+  /* @var \Drupal\node\Entity\Node $node */
+  $node = \Drupal::entityTypeManager()->getStorage('node')->load(123);
+  $entity_type = $node->getEntityType();
+  $entity_type->getSingularLabel();
+}
```
damontgomery, line 17: These is failing the automated test, because this "after" file needs to match exactly what Drupal Rector produces. I usually copy and past the "before" and then run Rector on it.

Can you update this one that way?

damontgomery (changes requested):
Thanks, @CocoaBean5!

I think you've found a limitation / but with our code as you mentioned. Can you see the notes about the failing tests and submit that as this PR and we can merge?

Can you submit an issue here with comments about the commented out code and we can try to resolve that issue and update this rule?

rector_examples/entity_type_getsingularlabel.php
damontgomery: After I wrote this, I suggested leaving it commented out if we don't want to address the rector issue right away.

rector_examples_updated/entity_type_getsingularlabel.php
CocoaBean5: @damontgomery The tests are passing now.

deprecation-index.yml
```diff
@@ -253,3 +253,8 @@
   PHPStan: 'Call to deprecated constant DATETIME_STORAGE_TIMEZONE: Deprecated in drupal:8.5.0 and is removed from drupal:9.0.0. Use Drupal\datetime\Plugin\Field\FieldType\DateTimeItemInterface::STORAGE_TIMEZONE instead.'
   Examples:
     - datetime_storage_timezone.php
+'EntityType::getLowercaseLabel()':
```
damontgomery, line 256: @CocoaBean5, I noticed that this rule is being named after the replacement and our rules are so far named after the things being replaced.

damontgomery:
Thanks, @CocoaBean5. Updated the names and merged #115.

PR #114, opened 2020-06-29
getStorage() rector on deprecated EntityManager

CocoaBean5:
## Description

- Call to deprecated method getStorage() of class Drupal​\​Core​\​Entity​\​EntityManager. Deprecated in drupal:8.0.0 and is removed from drupal:9.0.0. Use Drupal​\​Core​\​Entity​\​EntityTypeManagerInterface::getStorage() instead.

## Relevant Information

- Change record: https://www.drupal.org/node/2549139
- Method chaining issue is relevant on this PR


damontgomery:
EntityManager has an existing rector rule. I think the Acquia listing is just not displaying this.

See `entity_manager.php` for examples of what is covered.

I copied parts of your examples here and this is what gets replaced:

```
 function simple_example_2() {
-  $entity_manager = \Drupal::entityManager();
+  // TODO: Drupal Rector Notice: Please delete the following comment after you've made any necessary changes.
+  // We are assuming that we want to use the `entity_type.manager` service since no method was called here directly. Please confirm this is the case. See https://www.drupal.org/node/2549139 for more information.
+  $entity_manager = \Drupal::service('entity_type.manager');
   $node_storage = $entity_manager->getStorage('node');
   $node = $node_storage->load(123);

-  $node = \Drupal::entityManager()->getStorage('node')->load(123);
+  $node = \Drupal::service('entity_type.manager')->getStorage('node')->load(123);
 }
```

The edge cases as described in `src/Rector/Deprecation/EntityManagerRector.php` are around the cases where we need to replace `EntityManager` with classes other than `EntityTypeManager`.

For the chaining issue, this is resolved with #116.

PR #118, opened 2020-07-27
Upgrading dependency version of jawira/case-converter.

smrutha28:
Upgrading dependency version of jawira/case-converter

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
         "ast"
     ],
     "require": {
-        "jawira/case-converter": "^1.1",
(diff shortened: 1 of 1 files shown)


dwaghmare:
hi @damontgomery can  you give a thought on this https://github.com/palantirnet/drupal-rector/pull/118/files#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780R13 
I'm facing similar issue

damontgomery:
I removed the dependency in #119. I couldn't find any evidence of it being used and it had been copied from a repo this project was forked from.

PR #119, opened 2020-08-17
Remove Case Converter dependency since we aren't actually using it?

damontgomery:


damontgomery:
Closes #118

damontgomery:
I couldn't find any evidence of it being used and it had been copied from a repo this project was forked from.

damontgomery:
This is where it was used, https://github.com/drupal8-rector/drupal8-rector/blob/609820fd4213a62c6d0743d1253658b5ef33c682/src/Rector/CodeStyle/PropertyNameVariableNameRector.php.

The upstream project was doing more things than just handling deprecations, but we removed all of those features.

PR #121, opened 2020-09-11
Add rector for entity_delete_multiple

webflo:


webflo:
The code style of the generated code is not perfect. I did want to introduce variable to avoid conflicts with existing variables.

damontgomery:
Thanks, @webflo!

This looks good so far. I'll probably have more time to check it out maybe next week.

Would you be able to review https://github.com/palantirnet/drupal-rector#adding-a-rector-rule and add an example file and item in the index file?

The automated tests should check that example and it's helpful to understand your comments about the generated code too. :)

Thanks again.

webflo:
Added the required examples for the tests. Please review. Thanks!

damontgomery:
Sorry for the late reply. Other things came up and I wanted to apologize and provide an update. Not sure when we / I will be able to have a look at this.

damontgomery:
Thanks! The code is clear and the tests are passing. Sorry there was such a delay in reviewing this work.

I think this is ready to merge, but our automated tests are having authentication issues between Composer and Github that we will need to resolve and I think that may be something we want to fix before merging.

webflo:
Hi @damontgomery. I looked into the CI/composer issue and created #129

damontgomery:
Thanks @webflo!

PR #122, opened 2020-09-12
Upgrade to Rector 0.8

TomasVotruba:
Hi folks,

in recent 2 months there've been many changes in Rector, like moving from [YAML to PHP](https://getrector.org/blog/2020/08/31/rector-is-moving-from-yaml-to-php-configs-what-changes-and-how-to-get-ready).

I was thinking it might you save you some troubles, if I'd handle the upgrade myself, as I made those BC breaks you'd hate to fix :smile: 

So here it is


composer.json
```diff
@@ -10,7 +10,9 @@
         "ast"
     ],
     "require": {
-        "rector/rector-prefixed": "0.7.29"
+        "php": "^7.2",
```
TomasVotruba, line 13: PHP version must be part of `composer.json`, so composer can work with direct minimal version of this package.

This is minimal requirement of Rector 7.x

damontgomery: That makes sense, thanks.

composer.json
```diff
@@ -10,7 +10,9 @@
...
+        "jawira/case-converter": "^1.1",
```
damontgomery, line 14: This was removed since we weren't using it any more. We'll have to remove it here as well.

config/drupal-8/drupal-8.4-deprecations.yml
```diff
@@ -1,5 +0,0 @@
-# Contains automatic fixes for some deprecations introduced in Drupal 8.4.
```
damontgomery, line 1: Does this one have a `.php` version?

rector-config-docroot-dir.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+use Rector\Core\Configuration\Option;
+use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
+
+return static function (ContainerConfigurator $containerConfigurator): void {
+    $containerConfigurator->import(__DIR__ . '/vendor/palantirnet/drupal-rector/config/drupal-8/drupal-8-all-deprecations.yml');
```
damontgomery, line 9: Does this need to be updated from `.yml` to `.php`?

I think we'll want to also add commented out lines for the individual files in case people want to pick and choose.

rector-config-docroot-dir.php
```diff
@@ -0,0 +1,21 @@
...
+
+    $parameters = $containerConfigurator->parameters();
+
+    $parameters->set(Option::AUTOLOAD_PATHS, ['docroot/core', 'docroot/core/modules', 'docroot/modules', 'docroot/profiles']);
+    $parameters->set(Option::EXCLUDE_PATHS, ['*/upgrade_status/tests/modules/*']);
```
damontgomery, line 14: We'll want to add another commented out line here for people to exclude the test folders.

rector-config-docroot-dir.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+use Rector\Core\Configuration\Option;
+use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
+
+return static function (ContainerConfigurator $containerConfigurator): void {
```
damontgomery, line 8: Overall, we'll want to add back the comments to help people customize this file.

damontgomery (changes requested):
Thanks so much, Tomas!

We were looking into these updates a bit ago and this is going to help us a lot in upgrading to the new system.

I left some comments on some things that we'll want to add in before merging.

I think with the number of changes, it's probably easier for me to make a fork of your branch and work on those and then merge it in. Does that sounds like a good approach to you?

We may get a chance to work on this later in the week.

Thanks again.

TomasVotruba:
> I think with the number of changes, it's probably easier for me to make a fork of your branch and work on those and then merge it in. Does that sounds like a good approach to you?

Looking at the comments, I agree with you. Go for it!

I'll save my energy for Rector BC breaks ;)

rector-config-docroot-dir.php
TomasVotruba: Yes, if it's a Rector config.

config/drupal-8/drupal-8.4-deprecations.yml
TomasVotruba: I might have missed that, during rebase or my for to your master.

Could you add it?

composer.json
TomasVotruba: Sure, go for it.

I think I messed rebase, that's why it's here

TomasVotruba:
@damontgomery Btw, this might be interesting to you to prevent similar "upgrade" PRs in the future: https://github.com/rectorphp/rector/issues/4222

damontgomery:
Sorry for the late reply. Other things came up and I wanted to apologize and provide an update. Not sure when we / I will be able to have a look at this.

It's at the top of our list of TODOs.

TomasVotruba:
Surpased by https://github.com/palantirnet/drupal-rector/pull/134/files

PR #127, opened 2020-11-22
Adding db_drop_table rector rule.

prethiee:
Screenshot from 2020-11-22 22-54-41
Please review


damontgomery:
Thanks!

The code looks good. 

I think this is ready to merge, but our automated tests are having authentication issues between Composer and Github that we will need to resolve and I think that may be something we want to fix before merging.

damontgomery:
Thanks again!

I think this should pass. @webflo was able to resolve the automated test issue. Not sure why it won't pass here with the old error.

damontgomery:
@prethiee, sorry for merging this too quickly. We were having issues with CI server and I missed that this doesn't quite apply. Can you take a look at #130 where I reverted the code and provided a comment?

PR #129, opened 2020-12-03
Fix CI, remove git repo requirement

webflo:
I looked into the CI setup and could not understand why the the custom repository is required. rector-prefixed is available on Packagist. The GitHub Archive URLs have no rate-limit, therefor setting up authentication is not required.


damontgomery:
Thanks! This must have been left over from before it got added to packagist.

PR #130, opened 2020-12-03
Revert "Adding db_drop_table rector rule."

damontgomery:
Reverts palantirnet/drupal-rector#127

This is not quite ready. Sorry for merging before doing enough testing.

This is producing `\Drupal::database()->drop_table('user');` instead of `\Drupal::database()->schema()->dropTable('user');` with `->schema()`.

This will need more work since the base class doesn't take into account the pattern with a method in the middle.


prethiee:
Hey @damontgomery, thanks for reviewing and letting me know about this. I will take a look into this and see what I could do.

PR #134, opened 2021-03-24
Rector 0.10

mglaman:
Fixes #133
Fixes #135

## Remaining items

- [ ] Find out why parsing `src/Plugin/WebformElement/WebformLikert` crashed PHPStan's doc parser (This should be a follow up, it's not a Rector or this package bug)
- [x] Do not assume PHPUnit 7 in trait compatibility 

## Required follow ups

These are issues that need a fast follows once we merge this PR, otherwise the PR will just grow and grow and grow.

-  #138 Constants defined in .module files are not discovered
- Drupal 9 does not have PhpUnitCompatibility test traits (see https://github.com/palantirnet/drupal-rector/pull/134#issuecomment-821551267)
- Messaging that users should upgrade to 8.9 if on 8.7, 8.8, etc (see https://github.com/palantirnet/drupal-rector/pull/134#issuecomment-821547909)

## How to test

Add fork to your Drupal **8** site (assuming Composer build) composer.json
```
cd /path/to/drupal/recommended-project
composer config repositories.1 vcs https://github.com/bluehorndigital/drupal-rector.git
```

Require it
```
composer require palantirnet/drupal-rector:dev-rector-0.10
```

Configure it
```
cp vendor/palantirnet/drupal-rector/rector.php .
```

Run it!

Test it on your own code

```
php vendor/bin/rector process web/modules/custom
```

**OR** Grab the examples
```
cp -r vendor/palantirnet/drupal-rector/rector_examples web/modules/
php vendor/bin/rector process web/modules/rector_examples --dry-run
```

Verify it!

## Notes
Takes #122 commits and updates to 0.10

- composer: update rector to 0.8
- switch configs from YML to PHP
- drop duplicated info in editorconfig
- Add phpstan and rector dev trunk
- Add back EntityDeleteMultipleRector
- Update rule definitions for Rector 0.10

PR #143, opened 2021-04-28
Add support for entity_view

steveworley:
`entity_view` was [deprecated in 8.0.0](https://api.drupal.org/api/drupal/core%21includes%21entity.inc/function/entity_view/8.8.x).

Some complexities with the change in API as moving to this no longer supports the `$reset` parameter, but this will convert `entity_view` to a Drupal 9 compatible expression.


rector_examples/entity_view.php
```diff
@@ -0,0 +1,33 @@
...
+ */
+function simple_example() {
+  $entity = new stdClass();
+  $view = entity_view($entity, 'default');
+}
+
+function lagncode_example() {
+  $entity = new stdClass();
+  $langcode = 'de';
+  $veiw = entity_view($entity, 'default', $langcode);
```
damontgomery: minor: spelling of `$view`.

rector_examples/entity_view.php
```diff
@@ -0,0 +1,33 @@
...
+}
+
+function reset_cache_example() {
+  // Not supported.
```
damontgomery, line 25: Thanks for calling this out.

src/Rector/Deprecation/Base/EntityViewBase.php
```diff
@@ -0,0 +1,76 @@
...
+  public function refactor(Node $node): ?Node
+  {
+
+    if ($this->getName($node->name) !== 'entity_view') {
+      return NULL;
+    }
+
+    $name = new Node\Name\FullyQualified('Drupal');
+
+    $entity_type_method_name = new Node\Identifier('entityTypeManager');
```
damontgomery: Swap this for `$entityTypeManager_method_name`? like the ones below?

src/Rector/Deprecation/Base/EntityViewBase.php
```diff
@@ -0,0 +1,76 @@
...
+
+    $var = new Node\Expr\StaticCall($name, $entity_type_method_name);
+
+    $getViewBuilder_method_name = new Node\Identifier('getViewBuilder');
+
+    $entityRef = $node->args[0]->value;
```
damontgomery: Could we make this `$entity_reference`? I like how the names you've used are mostly matching the method names like `$getViewBuilder_method_name` which is very clear to me. :)

src/Rector/Deprecation/Base/EntityViewBase.php
```diff
@@ -0,0 +1,76 @@
+<?php
+
+namespace DrupalRector\Rector\Deprecation\Base;
+
+use DrupalRector\Utility\AddCommentTrait;
+use PhpParser\Node;
+use Rector\Core\Rector\AbstractRector;
+
+/**
+ * Replaced deprecated entity_load() calls.
```
damontgomery: Can we combine `EntityViewBase.php` and `EntityViewRector`? It looks like there was a base class, which we could refactor later, but for now, putting them together makes more sense to me while reading. The comments seem like they need to be updated too.

src/Rector/Deprecation/Base/EntityViewBase.php
```diff
@@ -0,0 +1,76 @@
...
+ * - Static replacement
+ * - The reset parameter is handled with a ternary operator
+ *
+ * Improvement opportunities
+ * - Dependency injection
+ */
+abstract class EntityViewBase extends AbstractRector
+{
+
+  use AddCommentTrait;
```
damontgomery: Did you want to add a comment to the replacement about checking for the `reset` option?

If not, we could remove this trait.

damontgomery (changes requested):
Thanks, @steveworley! And, hello!

I left a few comments. This looks close.

There are three other requirements for adding new rules
- Adding the new rector rule to the configuration file, config/drupal-8/drupal-8.0-deprecations.yml
- Updating the index (deprecation-index.yml)
- Adding the after example like rector_examples_updated/entity_create.php.

Also, there is a fairly large refactor going on to support newer versions of Rector. Because of that, I'm hesitant to merge new rules over the next week or so. If you want to build off that branch, we're in the final stages of moving in that direction. #134. This code should mostly work, but the configuration files have moved from YAML to php and some of the `use` statements have changed. Example: https://github.com/palantirnet/drupal-rector/pull/134/files#diff-069c02b1633ca197bab8d62713679cae3a5a80118c3b0a4ee51100c749ae6360.

Sorry again for the pieces moving from under you and happy to help take a look or make the updates after that PR is merged if you don't want to or don't have the time to make the updates.

steveworley:
Thanks for the review @damontgomery! Fancy seeing you here :)

I updated all the points and rebased against the updated rector branch let me know if you need anything else changed.

damontgomery:
:wave: I think reasonably I can take a look next week.

Thanks for the updates!

src/Rector/Deprecation/EntityViewRector.php
```diff
@@ -0,0 +1,95 @@
...
+     */
+    public function refactor(Node $node): ?Node
+    {
+        if ($this->getName($node->name) !== 'entity_view') {
+            return NULL;
+        }
+
+        $name = new Node\Name\FullyQualified('Drupal');
+
+        $entityTypManager = new Node\Identifier('entityTypeManager');
```
damontgomery, line 66: minor: spelling of `$entityTypeManager` missing the `e` in `Type`.

damontgomery (approved):
Thanks for the updates. The code is clear! The automated tests pass.

I would be happy to have this merged. Not sure if we should wait until / if we merge `rector-0.10-minimal-changes`.

PR #147, opened 2021-05-04
Add ability to fix Twig deprecations, with `replace` filter as first example

yusufhm:
The character by character replacement syntax for the `replace` Twig filter was [deprecated in 8.0.0](https://api.drupal.org/api/drupal/vendor%21twig%21twig%21lib%21Twig%21Extension%21Core.php/function/twig_replace_filter/8.0.x).

This PR creates a Twig file processor as well as a transformer for the filter.

### DIFF
--- a/src/Twig/Transformer/TwigReplaceTransformer.php
+++ b/src/Twig/Transformer/TwigReplaceTransformer.php
@@ -1,0 +1,32 @@
+<?php
+
+namespace DrupalRector\Twig\Transformer;
+
+use Rector\Core\Provider\CurrentFileProvider;
+use Symplify\RuleDocGenerator\ValueObject\CodeSample\CodeSample;
+use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
+
+final class TwigReplaceTransformer implements TwigTransformer
+{
+
+    public function getRuleDefinition(): RuleDefinition
+    {
+        return new RuleDefinition('Convert replace function into array', [
+            new CodeSample(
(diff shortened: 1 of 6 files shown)


damontgomery:
This is very cool! Thanks.

I may have time next week to try it out.

Community testing is appreciated!

src/Twig/Transformer/TwigReplaceTransformer.php
```diff
@@ -0,0 +1,32 @@
...
+                <<<'CODE_SAMPLE'
+{% set themeclass = item.title|replace({' ': '-'}) %}
+CODE_SAMPLE
+            ),
+        ]);
+    }
+
+    public function transform(string $twigContent): string
+    {
+        $twigContent = preg_replace('/replace\([\'"]{1}(.*)[\'"]{1},\s*[\'"]{1}(.*)[\'"]{1}\)/i', 'replace({\'$1\': \'$2\'})', $twigContent);
```
damontgomery, line 29: I'm not sure how I feel about this being handled by a complex regex statement.

Is there support in Rector to parse twig?

It could be helpful to parse these twig deprecations, but the power of Rector seems to be the ability to avoid regex in favor of these parsing libraries that are more reliable?

damontgomery:
Thanks again, @yusufhm!

Sorry for the late response. I'm not sure how to move forward here and would appreciate community feedback around the pro's and con's of parsing twig.

agentrickard:
This needs to be re-filed against `main`, sorry. @yusufhm

PR #148, opened 2021-05-21
Add rule for node_load_multiple

jedihe:
node_load_multiple() can be properly refactored by tweaking the existing logic in EntityLoadBase.

I tried improving it by using a protected variable ($isMultiple), defaulting it to FALSE in EntityLoadBase and setting it to TRUE in NodeLoadMultipleRector, as well as updating EntityLoadBase to now use it to determine $method_name, etc. For some reason, that change caused the is_null() check for $this->entityType to start returning TRUE, so I just left those changes out of the PR.

I also ran a quick sanity check by diffing a processed copy of rector_examples/node_load.php vs. the _updated version, which showed no changes at all.

### DIFF
--- a/src/Rector/Deprecation/NodeLoadMultipleRector.php
+++ b/src/Rector/Deprecation/NodeLoadMultipleRector.php
@@ -1,0 +1,42 @@
+<?php
+
+namespace DrupalRector\Rector\Deprecation;
+
+use DrupalRector\Rector\Deprecation\Base\EntityLoadBase;
+use Rector\Core\RectorDefinition\CodeSample;
+use Rector\Core\RectorDefinition\RectorDefinition;
+
+/**
+ * Replaced deprecated node_load_multiple() calls.
+ *
+ * See https://www.drupal.org/node/2266845 for change record.
+ *
+ * What is covered:
+ * - See EntityLoadBase.php
+ *
+ * Improvement opportunities
+ * - See EntityLoadBase.php
+ */
+final class NodeLoadMultipleRector extends EntityLoadBase
+{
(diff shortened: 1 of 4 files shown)


src/Rector/Deprecation/NodeLoadMultipleRector.php
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * See https://www.drupal.org/node/2266845 for change record.
+ *
+ * What is covered:
+ * - See EntityLoadBase.php
+ *
+ * Improvement opportunities
+ * - See EntityLoadBase.php
+ */
+final class NodeLoadMultipleRector extends EntityLoadBase
```
damontgomery, line 20: This seems like it will duplicate / be redundant to the implementation of https://github.com/palantirnet/drupal-rector/blob/master/src/Rector/Deprecation/NodeLoadRector.php.

damontgomery (commented):
Thanks for the update!

I'm not sure how to move forward. I think it's a good addition to support the multiple loading.

One of the guiding principles of the project is to have each class do a single thing and this introduces a class / set of classes that does two things. I think that could lead to confusion. It avoids copy & paste, but adds some complexity. It's not a lot, but I would tend towards creating an `EntityLoadMultipleBase` alongside `EntityLoadBase`.

The `NodeLoadMultipleRector` class is an example of this confusion. I think it duplicates `NodeLoadRector`, so there are two classes that do the same thing, which could be confusing.

Aside: We also need to update the index file `deprecation-index.yml`.

I want to be clear that I'm not saying these are requirements. They are suggestions. :)

Thanks again and happy to have other community feedback.

PR #153, opened 2021-06-16
Issue #3217942: Move to Rector 0.11

mglaman:
Moves from rector-prefixed to just rector

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
         "ast"
     ],
     "require": {
-        "rector/rector-prefixed": "~0.10.0",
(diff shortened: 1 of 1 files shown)


agentrickard:
@mglaman Right now, the lockfile is omitted from git. Do we want to add that?

agentrickard:
Other than the lockfile question, this looks good and tests runs as expected using:

`vendor/bin/rector process web/modules/custom/rector_examples --dry-run`

mglaman:
@agentrickard nah, we do not need the .lock. Generally, PHP libraries exclude it. Since it has no effect on the root project

PR #158, opened 2021-06-23
Revert "Issue #3217942: Move to Rector 0.11"

mglaman:
Reverts palantirnet/drupal-rector#153

Due to problems that @agentrickard found, we're reverting until we can have better test coverage.


agentrickard:
This is working as expected on DDEV.

PR #160, opened 2021-06-23
Issue #3217424 by mglaman: Add PHPUnit testing

mglaman:
## Description
Adding PHPUnit for stability and wins!

## To Test

Clone locally. Run phpunit

## Drupal.org issue
https://www.drupal.org/project/rector/issues/3217424

PR #161, opened 2021-06-24
Issue #3217942: Bump Rector to 0.11

mglaman:
## Description
Get to the latest Rector minor.

## To Test
Check out this PR and run against a Drupal site, verify 0.11 works as expected.

## Drupal.org issue
https://www.drupal.org/project/rector/issues/3217942

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
         "ast"
     ],
     "require": {
-        "rector/rector-prefixed": "~0.10.0",
+        "rector/rector": "~0.11.0",
         "webflo/drupal-finder": "^1.2"
     },
     "license": "MIT",


mglaman:
@agentrickard I re-tried this for kicks, due to PHPUnit being buggy and wondering if we'll have improvements on 0.11.0

The problem.. I can't reproduce or earlier error inside of DDEV :(

mglaman:
Bumping to 0.11.0 _does_ fix the weird PHPUnit autoloading errors: https://github.com/palantirnet/drupal-rector/pull/160

I wonder if the DDEV error is only reproducible when using Docker for Mac? I mean, I wouldn't be surprised :(

agentrickard:
I was planning to test this one and then #160.

I got everything to install correctly after deleting the `vendor` directory and `composer.lock` file, then using this issue as a require statement.

agentrickard:
Runs correctly on Drupal 8, as expected.

PR #163, opened 2021-07-07
Issue #3222671: AssertLegacyTrait rector rules

mglaman:
## Description
Adds rules for legacy assertion methods. This takes care of the top 7 (excluding `drupalPostForm` which was already fixed.)

* assertText
* assertEqual
* assertIdentical
* assertResponse
* assertRaw
* assertFieldByName 

## To Test

## Drupal.org issue
https://www.drupal.org/project/rector/issues/3222671

### DIFF
--- a/src/Rector/Deprecation/AssertEqualRector.php
+++ b/src/Rector/Deprecation/AssertEqualRector.php
@@ -15,7 +15,7 @@ final class AssertEqualRector extends AssertLegacyTraitBase
 
     public function getRuleDefinition(): RuleDefinition
     {
-        return new RuleDefinition('Fixes deprecated UiHelperTrait::drupalPostForm() calls', [
+        return new RuleDefinition('Fixes deprecated AssertLegacyTrait::assertEqual() calls', [
             new CodeSample(
                 <<<'CODE_BEFORE'
 $this->assertEqual('Actual', 'Expected', 'Message');
--- a/src/Rector/Deprecation/AssertEqualRector.php
+++ b/src/Rector/Deprecation/AssertEqualRector.php
@@ -1,0 +1,31 @@
+<?php declare(strict_types=1);
+
+namespace DrupalRector\Rector\Deprecation;
+
+use DrupalRector\Rector\Deprecation\Base\AssertLegacyTraitBase;
+use Symplify\RuleDocGenerator\ValueObject\CodeSample\CodeSample;
+use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
+
+final class AssertEqualRector extends AssertLegacyTraitBase
+{
+
+    protected $deprecatedMethodName = 'assertEqual';
+    protected $methodName = 'assertEquals';
+    protected $isAssertSessionMethod = false;
+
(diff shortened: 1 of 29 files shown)


src/Rector/Deprecation/AssertEqualRector.php
```diff
@@ -0,0 +1,31 @@
...
+final class AssertEqualRector extends AssertLegacyTraitBase
+{
+
+    protected $deprecatedMethodName = 'assertEqual';
+    protected $methodName = 'assertEquals';
+    protected $isAssertSessionMethod = false;
+
+    public function getRuleDefinition(): RuleDefinition
+    {
+        return new RuleDefinition('Fixes deprecated UiHelperTrait::drupalPostForm() calls', [
```
mglaman: Wrong rule definition statement, fixes `assertEquals`

src/Rector/Deprecation/AssertTextRector.php
```diff
@@ -0,0 +1,45 @@
...
+
+    // @codingStandardsIgnoreLine
+    protected $comment = 'Verify the assertion: pageTextContains() for HTML responses, responseContains() for non-HTML responses.' .
+        PHP_EOL . '// The passed text should be HTML decoded, exactly as a human sees it in the browser.';
+    protected $deprecatedMethodName = 'assertText';
+    protected $methodName = 'pageTextContains';
+
+    public function getRuleDefinition(): RuleDefinition
+    {
+        return new RuleDefinition('Fixes deprecated UiHelperTrait::drupalPostForm() calls', [
```
mglaman: Wrong words

agentrickard:
I tested this locally in 3 separate environments, and the PHPUnit tests always fail with:

```
1) DrupalRector\Tests\Rector\Deprecation\AssertFieldByNameRectorTest\AssertFieldByNameRectorTest::test with data set #0 (RectorPrefix20210706\Symplify\SmartFileSystem\SmartFileInfo Object (...))
PHPStan\BetterReflection\Reflector\Exception\IdentifierNotFound: PHPStan\BetterReflection\Reflection\ReflectionClass "Drupal\Tests\BrowserTestBase" could not be found in the located source
```

The easiest to replicate is by checking out this PR into https://github.com/palantirnet/drupal-rector-sandbox/tree/php8, which installs PHP 8.0, Drupal 9.2, PHPUnit 9.5.6, and phpstan 0.12.91.

mglaman:
@agentrickard odd, it's as if it isn't picking up the stub file (added in the last PR)

I'm referring to this: https://github.com/palantirnet/drupal-rector/blob/main/stubs/Drupal/Tests/BrowserTestBase.php

agentrickard:
I can confirm that this testing error is introduced in this PR. I think it’s a conflict when Drupal core is installed in the project, which is how I have been testing everything.

Reverting the sandbox to `main` fixes the issue.

It may be that I am testing *incorrectly* and that the PHPUnit tests should not be run when Drupal is present -- I don't see it installed in https://github.com/palantirnet/drupal-rector/pull/163/checks?check_run_id=3068426012 .

Ideally, however, we could run tests when Drupal is present.

PR #170, opened 2021-08-12
Adds AssertLegacy and other new items to the index.

agentrickard:
## Description

Adds items to the deprecation-index.yml file that are needed for Update Status.

## To Test
- Needs to be tested as a replacement for the patch in https://www.drupal.org/project/upgrade_status/issues/3224920

## Drupal.org issue

https://www.drupal.org/project/rector/issues/3224919

## Open question

- [ ] How do we document examples in the D9 version of the codebase? Do we need to?

PR #173, opened 2021-08-18
Use `"` when `'` is in the deprecation message.

mglaman:
Fixes YAML failures in https://github.com/palantirnet/drupal-rector/pull/170

### DIFF
--- a/deprecation-index.yml
+++ b/deprecation-index.yml
@@ -434,7 +434,7 @@
     Examples:
 'AssertLegacyTrait::getAllOptions()':
     Rector: GetAllOptionsRector.php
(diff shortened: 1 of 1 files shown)


agentrickard:
I'm going to re-roll that PR with the new PHPStan stuff from https://www.drupal.org/project/rector/issues/3224919#comment-14193202 and fix it there.

agentrickard:
These are all in Gábor Hojtsy's list of references not found in PHPStan because there are no instances in contrib.

Apparently, PHPStan messages don't include parameters in this way.

agentrickard:
@mglaman Updated the index and merged in the `push` fix. Ready for review.

PR #174, opened 2021-08-18
Only load stubs in dev

mglaman:
## Description
Prevents stubs used for testing from breaking end user's PHPUnit tests for Drupal

## To Test
- Checkout
- Run a Functional test in core. No errors

## Drupal.org issue
https://www.drupal.org/project/rector/issues/3223276

PR #175, opened 2021-08-18
Make PassRector more specific and fix documentation

mglaman:
## Description
- Add to deprecation-index and fix definition doc
- Issue #3228113: Make PassRector more specific and fix documentation

## To Test

## Drupal.org issue
https://www.drupal.org/project/rector/issues/3228113

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,15 +35,15 @@
     "autoload": {
         "psr-4": {
             "DrupalRector\\": "src"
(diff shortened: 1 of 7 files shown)


mglaman:
This has the fix from https://github.com/palantirnet/drupal-rector/pull/174, so if this merges first that one is not needed.

PR #177, opened 2021-08-18
Issue #3226796: Include rector-symfony config set lists for deprecated code fixes

mglaman:
## Description
Adds all of the Symfony 4 Rector rules for deprecated code.

## To Test
Functional tests shows deprecated class renames and fixing of the `dispatch` call (object first, name second and optional).

## Drupal.org issue
https://www.drupal.org/project/rector/issues/3226796

### DIFF
--- a/config/drupal-9/drupal-9.0-deprecations.php
+++ b/config/drupal-9/drupal-9.0-deprecations.php
@@ -3,8 +3,14 @@
 declare(strict_types=1);
 
 use Rector\PHPUnit\Set\PHPUnitSetList;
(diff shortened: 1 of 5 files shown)


mglaman:
Failures are due to https://github.com/palantirnet/drupal-rector/pull/183

catch56:
The rename for Symfony Contracts event is going to conflict with Drupal's forward compatibility layer for that class - ideally we want to rename to the new Drupal Event class which bridges Symfony 4-6 rather than rename to Symfony which will be a hard break with Symfony 3. See https://www.drupal.org/node/3159012 for details.

PR #183, opened 2021-09-02
Issue #3231016: Access level must be protected

FlorentTorregrosa:
## Description

Follow up of changes in Rector.

## To Test

Run Rector on a code base.

## Drupal.org issue

https://www.drupal.org/project/rector/issues/3231016

### DIFF
--- a/src/Rector/Deprecation/AssertNoUniqueTextRector.php
+++ b/src/Rector/Deprecation/AssertNoUniqueTextRector.php
@@ -17,11 +17,6 @@ final class AssertNoUniqueTextRector extends AbstractRector
 
     use GetDeclaringSourceTrait;
 
-    /**
-     * @todo remove when property is no longer private in AbstractRector.
-     */
-    protected $nodesToAddCollector;
-
     public function __construct(
         NodesToAddCollector $nodesToAddCollector
     ) {
--- a/src/Rector/Deprecation/AssertNoUniqueTextRector.php
+++ b/src/Rector/Deprecation/AssertNoUniqueTextRector.php
@@ -17,12 +17,6 @@ final class AssertNoUniqueTextRector extends AbstractRector
 
     use GetDeclaringSourceTrait;
 
-    public function __construct(
-        NodesToAddCollector $nodesToAddCollector
-    ) {
-        $this->nodesToAddCollector = $nodesToAddCollector;
-    }
-
     public function getRuleDefinition(): RuleDefinition
     {
         return new RuleDefinition('Fixes deprecated AssertLegacyTrait::assertNoUniqueText() calls', [
--- a/src/Rector/Deprecation/AssertNoUniqueTextRector.php
+++ b/src/Rector/Deprecation/AssertNoUniqueTextRector.php
@@ -17,6 +17,11 @@ final class AssertNoUniqueTextRector extends AbstractRector
 
     use GetDeclaringSourceTrait;
 
+    /**
+     * @todo remove when property is no longer private in AbstractRector.
+     */
+    protected $nodesToAddCollector;
+
     public function __construct(
         NodesToAddCollector $nodesToAddCollector
     ) {
--- a/src/Rector/Deprecation/AssertNoUniqueTextRector.php
+++ b/src/Rector/Deprecation/AssertNoUniqueTextRector.php
@@ -17,11 +17,6 @@ final class AssertNoUniqueTextRector extends AbstractRector
 
     use GetDeclaringSourceTrait;
 
-    /**
-     * @todo remove when property is no longer private in AbstractRector.
-     */
-    private $nodesToAddCollector;
-
     public function __construct(
         NodesToAddCollector $nodesToAddCollector
     ) {
--- a/src/Rector/Deprecation/UiHelperTraitDrupalPostFormRector.php
+++ b/src/Rector/Deprecation/UiHelperTraitDrupalPostFormRector.php
@@ -12,11 +12,6 @@ use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
 final class UiHelperTraitDrupalPostFormRector extends AbstractRector
 {
 
-    /**
-     * @todo remove when property is no longer private in AbstractRector.
-     */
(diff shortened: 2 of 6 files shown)


salvamomo:
I've just come across this issue as well. This patch works for me :+1:

SerShevchyk:
Please approve this patch

mglaman:
Noice! They made it protected in the source?

mglaman:
Here's the commit which adjusted the property's visibility: https://github.com/rectorphp/rector-src/commit/c732f61317d64898b7b31ab171d7f1e52caca9f9. It does not look like it has been tagged yet, so we cannot merge. Or we have to keep our property definition, but change it to protected to be forward compatible

src/Rector/Deprecation/AssertNoUniqueTextRector.php
```diff
@@ -17,11 +17,6 @@ final class AssertNoUniqueTextRector extends AbstractRector
 
     use GetDeclaringSourceTrait;
 
-    /**
-     * @todo remove when property is no longer private in AbstractRector.
-     */
-    private $nodesToAddCollector;
-
```
mglaman, line 24: Keep but make protected since change is not released.

src/Rector/Deprecation/UiHelperTraitDrupalPostFormRector.php
```diff
@@ -12,11 +12,6 @@
 final class UiHelperTraitDrupalPostFormRector extends AbstractRector
 {
 
-    /**
-     * @todo remove when property is no longer private in AbstractRector.
-     */
-    private $nodesToAddCollector;
-
```
mglaman, line 19: Keep but make protected since change is not released.

mglaman (changes requested):

FlorentTorregrosa:
Hello,

Thanks everyone for the feedbacks. PR updated.

@mglaman do you want to mark the review points as resolved or should I do that? When I do code review I prefer that it is the reviewer that judge if it ok or not.

src/Rector/Deprecation/AssertNoUniqueTextRector.php
FlorentTorregrosa: Done

src/Rector/Deprecation/UiHelperTraitDrupalPostFormRector.php
FlorentTorregrosa: Done

agentrickard:
Looks like Rector 0.11.53 includes this change.

agentrickard:
Tests are green @mglaman This could user another review.

src/Rector/Deprecation/AssertNoUniqueTextRector.php
```diff
@@ -17,11 +17,6 @@ final class AssertNoUniqueTextRector extends AbstractRector
...
     public function __construct(
         NodesToAddCollector $nodesToAddCollector
     ) {
```
mglaman: We could probably drop this constructor if it is fixed upstream? Need to review full code

iyyappandrupal: Hi @mglaman What do you mean by fixed upstream? It is working well after removing of constructor.

mglaman: @iyyappandrupal upstream being the class in Rector. If it works with the constructor removed, the PR should be updated to remove it (cc @FlorentTorregrosa or @agentrickard) in both classes. That way we have less code to maintain.

FlorentTorregrosa: I am on it

iyyappandrupal: @mglaman Do you mean AbstractRector class in Rector library?Thanks

FlorentTorregrosa: PR updated. But warning, I have not tested locally as I was not on my standard dev environment.

iyyappandrupal: @FlorentTorregrosa I tested morning in my project after removing the constructor, it works well.

neclimdul:
worked for me. 

If I understand correctly this is working around an internal change upstream. Does it then make sense to require a minimum version for releases with this code? Seems that would force earlier releases for older rector installs and avoid people accidentally breaking their installs if they don't update all their dependencies.

mglaman (approved):
I'm marking it approved, but we probably should bump the `composer.json` to `~0.11.53`

Currently, it is `~0.11.39`. I honestly am not _too_ worried, because anyone updating the library should be doing so with `-W` for "with all dependencies" or they are probably updating with `composer update`.

I don't have access to edit the file to provide the suggestion. So @agentrickard feel free on merge, maybe? or a follow up

mglaman:
@agentrickard it would be good to also approve the PR for workflows

mglaman:
The tests are passing locally, were failing before PR
Screen Shot 2021-09-22 at 2 12 56 PM

agentrickard:
@mglaman When I bumped the rector version to 0.11.53, the functional tests started to fail.

mglaman:
@agentrickard The functional tests are broken because PHPUnit rules in Rector were updated to add no assert annotations automatically (kind of cool.)

```
+    /**
+     * @doesNotPerformAssertions
+     */
```

I would say add them as a fix because it's just new improvements from Rector's rules. I could make a new PR, but that would fail due to the `private` property and essentially duplicate this entire PR.

@FlorentTorregrosa Are you able to apply the fixes to the `rector_examples_updated` that are shown in this job run? https://github.com/palantirnet/drupal-rector/pull/183/checks?check_run_id=3701796890

Then we can hopefully get this merged and have drupal-rector humming again.

FlorentTorregrosa:
@mglaman ok, I am on it. I will rebase my PR against main.

FlorentTorregrosa:
PR updated :)

agentrickard:
Nice work everyone!

PR #190, opened 2022-02-09
Handle custom message types

mglaman:
## Description
Falls back to addMessage when an unknown message type is specified. Also fixes resolving the type from a variable in normal use cases!

## To Test

## Drupal.org issue
https://www.drupal.org/project/rector/issues/3241156

### DIFF
--- a/fixtures/d8/rector_examples_updated/drupal_set_message.php
+++ b/fixtures/d8/rector_examples_updated/drupal_set_message.php
@@ -39,8 +39,5 @@ function message_type_as_variable() {
 
   $type = 'warning';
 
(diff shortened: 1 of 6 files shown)


mglaman:
I need to fix the functional test fixtures for D8

```
     $type = 'warning';
 
-    // TODO: Drupal Rector Notice: Please delete the following comment after you've made any necessary changes.
-    // This needs to be replaced, but Rector was not yet able to replace this because the type of message was set with a variable. If you need to continue to use a variable, you might consider using a switch statement.
-    // @noRector
-    drupal_set_message($message, $type);
+    \Drupal::messenger()->addWarning($message);
   }
```

Because it's improved and fixed!

mglaman:
Updated fixtures to fix test failure.

PR #194, opened 2022-03-31
Disable AUTO_IMPORT_NAMES option

malcomio:
See https://www.drupal.org/project/rector/issues/3269329#comment-14463456


mglaman (approved):
Looks good to me! This is something phpcs can handle, I agree.

mglaman:
This is outdated with recent changes, and should be included in 0.13.0 and enhanced with https://github.com/palantirnet/drupal-rector/pull/209

PR #198, opened 2022-05-12
Issue #3280205: drupalPostForm() with 1st param NULL

claudiu-cristea:
## Description

See https://www.drupal.org/project/rector/issues/3280205

## To Test

See fixture from PR

## Drupal.org issue

https://www.drupal.org/project/rector/issues/3280205

### DIFF
--- a/src/Rector/Deprecation/UiHelperTraitDrupalPostFormRector.php
+++ b/src/Rector/Deprecation/UiHelperTraitDrupalPostFormRector.php
@@ -70,20 +70,24 @@ CODE_AFTER
         assert($node instanceof Node\Expr\MethodCall);
         if ($this->getName($node->name) === 'drupalPostForm') {
             [$path, $edit, $button, $options, $htmlId] = $this->safeArgDestructure($node);
-            $pathValue = $path->value;
-            assert($pathValue instanceof Node\Scalar\String_);
-            if ($options === null) {
-                $drupalGetNode = $this->nodeFactory->createLocalMethodCall('drupalGet', [$path]);
-            } else {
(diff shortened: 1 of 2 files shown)


src/Rector/Deprecation/UiHelperTraitDrupalPostFormRector.php
```diff
@@ -70,20 +74,24 @@ public function refactor(Node $node): ?Node
...
             } else {
                 $submitFormNode = $this->nodeFactory->createLocalMethodCall('submitForm', [$edit, $button, $htmlId]);
             }
-            $this->nodesToAddCollector->addNodeBeforeNode($drupalGetNode, $node);
+
+            $pathValue = $path->value;
+            if (!$pathValue instanceof Node\Expr\ConstFetch || strtolower((string) $pathValue->name) !== 'null') {
+                assert($pathValue instanceof Node\Scalar\String_);
+                if ($options === null) {
+                    $drupalGetNode = $this->nodeFactory->createLocalMethodCall('drupalGet', [$path]);
```
mglaman, line 88: So, I wrote bad code before by leaving that `assert` on `$pathValue`. It could be a variable referencing a string, and I don't think it would be a resolved value here.

What if we refactored this to:

```
$pathIsNull = $pathValue instanceof Node\Expr\ConstFetch && strtolower((string) $pathValue->name) === 'null';
if ($pathIsNull === false) {
```

Reading the above logic, the `||` is throwing me off, because the object may not have a `->name` property to resolve. That would require and `&&` asserting that `$pathValue` is `ConstFetch`.

mglaman: It works, I'll approve so we can get it fixed. Just my feedback.

mglaman (approved):

PR #202, opened 2022-05-25
Issue #3282217: file_build_uri() is deprecated

claudiu-cristea:
## Description

`file_build_uri()` is deprecated in Drupal 9.3. See https://www.drupal.org/node/3223091

## To Test

Run Rector on a code-base that includes `file_build_uri()`

## Drupal.org issue

https://www.drupal.org/project/rector/issues/3282217

### DIFF
--- a/src/Rector/Deprecation/FileBuildUrlRector.php
+++ b/src/Rector/Deprecation/FileBuildUrlRector.php
@@ -1,0 +1,82 @@
+<?php
+
+namespace DrupalRector\Rector\Deprecation;
+
+use PhpParser\Node;
+use Rector\Core\Rector\AbstractRector;
+use Symplify\RuleDocGenerator\ValueObject\CodeSample\CodeSample;
+use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
+
+final class FileBuildUrlRector extends AbstractRector
+{
+    /**
+     * @inheritdoc
+     */
+    public function getRuleDefinition(): RuleDefinition
+    {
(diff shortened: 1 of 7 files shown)


src/Rector/Deprecation/FileBuildUrlRector.php
```diff
@@ -0,0 +1,82 @@
+<?php
+
+namespace DrupalRector\Rector\Deprecation;
+
+use PhpParser\Node;
+use Rector\Core\Rector\AbstractRector;
+use Symplify\RuleDocGenerator\ValueObject\CodeSample\CodeSample;
+use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
+
+final class FileBuildUrlRector extends AbstractRector
```
mglaman: Shouldn't this be FileBuildUriRector not FileBuildUrlRector

mglaman (changes requested):
Rule name does not match function name it's `uri` not `url`

src/Rector/Deprecation/FileBuildUrlRector.php
claudiu-cristea: My bad, yes

claudiu-cristea:
@mglaman, I've renamed the class. I've also escalated the issue with useless parenthesis here https://github.com/nikic/PHP-Parser/issues/839. However, I have no idea if it's possible & how to use a different printer.

mglaman (approved):
Looks good to me!

PR #203, opened 2022-05-25
Remove remaining Rector conflicts up to 0.12.21

mglaman:
## Description
Removes remaining Rector conflicts, up to 0.12.21. Turns out the improvements to Rector config are not breaking changes. The reason we had failures was to:

* Removal of constant Option::IMPORT_DOC_BLOCKS 
* PHPUnit failed due to missing patches on Rector -dev dependencies
* ParameterProvider property removed from abstract rector rule base class.

Tried bumping to https://github.com/rectorphp/rector/releases/tag/0.12.22, but then we hit deprecated option calls and must update to the new configuration format. Ran out of time to work on that. Maybe this is the next best "hop" to allow Rector updates and include provided Rector rule fixes like https://github.com/palantirnet/drupal-rector/pull/202 and https://github.com/palantirnet/drupal-rector/pull/201

## To Test

## Drupal.org issue

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,11 +10,11 @@
         "ast"
     ],
     "require": {
-        "rector/rector": "~0.12.20",
+        "rector/rector": "~0.12.21",
         "webflo/drupal-finder": "^1.2"
     },
     "conflict": {
-        "rector/rector": ">0.12.20"
+        "rector/rector": ">0.12.21"
     },
     "license": "MIT",
     "authors": [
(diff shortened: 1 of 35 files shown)


composer.json
```diff
@@ -65,20 +65,24 @@
     },
     "minimum-stability": "dev",
     "prefer-stable": true,
+    "extra": {
+        "enable-patching-note": "We need the following to allow rector/rector-src to patch dependencies.",
+        "enable-patching": true
+    },
```
mglaman, line 71: Kept getting errors about RectorConfig and a generator being null. Turns out that Rector patches symfony/dependency-injecting and other items. This is fine for `rector/rector` but not when using dev dependencies. I reviewed `rector/rector-phpunit` and found this line. By adding it, `symfony/dependency-injection is patched properly to make our tests run.

**NOTE** This will not affect Drupal sites or patch a Drupal site's `symfony/dependency-injection. This only happens for `require-dev`, when running our own PHPUnit tests. It has no effect on the root Composer project.

src/Rector/Deprecation/Base/DBBase.php
```diff
@@ -49,6 +50,18 @@ abstract class DBBase extends AbstractRector
...
+    public function __construct(ParameterProvider $parameterProvider)
+    {
+        $this->parameterProvider = $parameterProvider;
+    }
+
+    protected function getParameterProvider(): ParameterProvider
+    {
+        return $this->parameterProvider;
+    }
+
```
mglaman: This is a lot of boilerplate and should really go into a shared service that all of the rules use. However, we delegate all ownership of configuration to end-users. So, this is a bit messy but there's no good workaround since the property was removed from AbstractRector base class.

This is used to check if `drupal_rector_notices_as_comments` has been flagged in the configuration:

```
$parameters->set('drupal_rector_notices_as_comments', true);
```

Part of me wonders if anyone ever sets that to false and if we should just default to `true` and add comments.

rector.php
```diff
@@ -28,8 +28,6 @@
     $parameters->set(Option::FILE_EXTENSIONS, ['php', 'module', 'theme', 'install', 'profile', 'inc', 'engine']);
     $parameters->set(Option::AUTO_IMPORT_NAMES, true);
     $parameters->set(Option::IMPORT_SHORT_CLASSES, false);
-    // @phpstan-ignore-next-line
-    $parameters->set(Option::IMPORT_DOC_BLOCKS, false);
```
mglaman, line 32: removed in 0.12.21: https://github.com/rectorphp/rector-src/commit/c35cc97d381f97c602cccb86ccf1d688a869cbdd

mglaman:
Latest failure:

```
         "DrupalRector\Rector\Deprecation\DBInsertRector": argument             
         "$parameterProvider" of method                                         
         "DrupalRector\Rector\Deprecation\Base\DBBase::__construct()" has type  
         "Symplify\PackageBuilder\Parameter\ParameterProvider" but this class   
         was not found.       
```

We're relying on an inherited dependency from Rector for \Symplify\PackageBuilder\Parameter\ParameterProvider. This is bundled in `rector/rector`, which means it is namespaced in its vendor packages. Locally, with `require-dev` we have the package available so it passes. 

See https://github.com/rectorphp/rector/blob/0.12.21/vendor/symplify/package-builder/src/Parameter/ParameterProvider.php

```
<?php

declare (strict_types=1);
namespace RectorPrefix20220418\Symplify\PackageBuilder\Parameter;
```

There's no way to re-use that class.

src/Rector/Deprecation/Base/DBBase.php
mglaman: Resolved by using ConfigurableRectorInterface and container parameter injection.

agentrickard:
@mglaman Do you think this warrants a release? Or is it incremental as we change Rector 13?

mglaman:
@agentrickard I think so. That way we have incremental releases as we edge closer.

Bumping to Rector [0.12.22](https://github.com/rectorphp/rector/releases/tag/0.12.22) raises configuration option deprecations, which means the next drupal-rector release after this will have breaking changes. This makes it easier for folks to upgrade, I think.

mglaman:
Fast follow: releasing 0.12.4 would include the following without bumping into breaking config changes (keeping it somewhat stable):

* https://github.com/palantirnet/drupal-rector/pull/201
* https://github.com/palantirnet/drupal-rector/pull/202

PR #215, opened 2022-11-09
Use this->t instead of t

chrfritsch:
## Description
It checks if a `t()` call can be converted to `$this->t()`

## To Test

## Drupal.org issue

### DIFF
--- a/src/Rector/BestPractice/UseThisTInsteadOfTRector.php
+++ b/src/Rector/BestPractice/UseThisTInsteadOfTRector.php
@@ -1,0 +1,80 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DrupalRector\Rector\BestPractice;
+
+use PhpParser\Node;
+use Rector\Core\Rector\AbstractRector;
+use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
+use Symplify\RuleDocGenerator\ValueObject\CodeSample\ConfiguredCodeSample;
+use PhpParser\Node\Expr\FuncCall;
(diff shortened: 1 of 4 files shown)


tobiasbaehr:
There is already a rule. Perhaps the rule can be reused.


https://github.com/rectorphp/rector/blob/main/docs/rector_rules_overview.md#funccalltomethodcallrector

src/Rector/BestPractice/UseThisTInsteadOfTRector.php
```diff
@@ -0,0 +1,80 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DrupalRector\Rector\BestPractice;
+
+use PhpParser\Node;
+use Rector\Core\Rector\AbstractRector;
```
nlighteneddesign:
```suggestion
use Rector\Rector\AbstractRector;
```

bbrala:
This rector needs rewriting. Parent is no longer available by default, so we need to rearcitect this, see https://getrector.com/blog/rector-017-using-scoped-traverse

think something like:

1. Add betternodefinder in constructor (`public function __construct(protected BetterNodeFinder $betterNodeFinder){}`)
2. Match on class method probably (`Node\Stmt\ClassMethod::class`)
3. Then use `$stmts = $this->betterNodeFinder->find($node, FuncCall::class);` i think to find if t() is used
4. Use the context of the class to find out if the class has a `$this->t(...)` availabe. This might neem swithcing with 3, it might be faster that way to first see if we have a t function available, then if so see if it is uses wrongly?
5. Make the change to the functioncall.

PR #219, opened 2023-03-13
Update rector config to resolve bootstrap issues with recent rector releases

goba:
## Description

See https://www.drupal.org/project/rector/issues/3346620, I reproduced the same problem locally and adding this conflict to my global composer file helped. Matt Glaman suggested to add it to drupal-rector which makes a ton of sense.

## To Test

Install drupal-rector as usual. It will produce the error from https://www.drupal.org/project/rector/issues/3346620 instead of doing anything useful. Apply this PR and it works (for me at least).

## Drupal.org issue

https://www.drupal.org/project/rector/issues/3346620

### DIFF
--- a/config/drupal-8/drupal-8-all-deprecations.php
+++ b/config/drupal-8/drupal-8-all-deprecations.php
@@ -27,7 +27,7 @@ return static function (RectorConfig $rectorConfig): void {
     // the following rule adds unneeded annotations.
     $rectorConfig->services()->remove(AddDoesNotPerformAssertionToNonAssertingTestRector::class);
 
-    $rectorConfig->bootstrapFiles([
+    $rectorConfig->autoloadPaths([
         __DIR__ . '/../drupal-phpunit-bootstrap-file.php'
     ]);
 };
--- a/composer.json
+++ b/composer.json
@@ -71,7 +71,7 @@
         "phpstan/extension-installer": "^1.1",
         "phpstan/phpstan": "^1.0",
         "phpstan/phpstan-deprecation-rules": "^1.0",
-        "phpunit/phpunit": "^9.5",
+        "phpunit/phpunit": "^10.0",
         "rector/rector-src": "dev-main",
         "symplify/vendor-patches": "^11.0",
         "symfony/yaml": "^5 || ^6"
--- a/composer.json
+++ b/composer.json
@@ -13,6 +13,9 @@
         "rector/rector": "~0.15.2",
         "webflo/drupal-finder": "^1.2"
     },
+    "conflict": {
+        "rector/rector": ">=0.15.19"
+    },
     "license": "MIT",
     "authors": [
         {
--- a/composer.json
+++ b/composer.json
@@ -13,9 +13,6 @@
         "rector/rector": "~0.15.2",
         "webflo/drupal-finder": "^1.2"
     },
-    "conflict": {
-        "rector/rector": ">=0.15.19"
-    },
     "license": "MIT",
     "authors": [
         {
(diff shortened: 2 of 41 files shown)


tobiasbaehr:
Lets try the comment of https://github.com/rectorphp/rector/issues/7795#issuecomment-1465995891

tobiasbaehr:
I updated rector to the latest version and the changes from the comment works.

goba:
Did those updates in this PR, let's see @tobiasbaehr

goba:
Looking at the fails, it seems like the first one is phpunit 9.5 requires sebastian/diff ^4.0.3, while rector/rector-src dev-main requires sebastian/diff ^5.0, so they cannot be installed at once. So appears like we need phpunit 10 compatibility first.

goba:
This one now passes! "drupal_rector_sandbox / Rector Sandbox | PHP 7.4 (pull_request)"

Next one is "phpunit / PHPUnit | PHP 8.1 (pull_request)" which fails due to lack of static data provider methods. PHPUnit 9.5 deprecated those and PHPUnit 10 wants to look for the static methods but they are not there. So many tests fails due to lack of test methods then.

goba:
The failing functional test is producing a bunch of these types of issues:

Could not process                                                      
         "/home/runner/drupal/web/modules/custom/rector_examples/tests/src/Funct
         ional/AssertFieldCheckedTest.php" file, due to:                        
         "System error: "Method assertFieldChecked() was not found in reflection
         of class                                                               
         Drupal\Tests\rector_examples\Functional\AssertFieldCheckedTest."

goba:
So now I have 3 of hte 6 checks green. *However* I think there is a block now with Drupal 10.1 not even allowing PHPUnit 10, so not sure how we could test drupal-rector on PHPUnit 10 with Drupal 8.9 and Drupal 9 :/ The error found above are I think due to the PHPUnit upgrader thingy not being able to deal with PHPUnit 10, so compatible base classes are not present.

So I am not sure we can update to PHPUnit 10 quite yet? https://www.drupal.org/project/drupal/issues/3217904 is the core issue to update core to support it and its not planned to be backported to Drupal 9.

Should we go on and add the conflict instead? We would need to go all the way back to https://packagist.org/packages/rector/rector-src#0.15.11 to have PHPUnit 9.5 support still on rector. Sounds like we may be stuck there to be Drupal 8/9 compatible?

goba:
Per @mglaman, this is not an incompatibility. Drupal doesn’t dump `Drupal\Tests\` into Composer. So it needs the "phpunit-bootstrap nonsense". I don't really know how to proceed on that though.

agentrickard:
The functional test runner is still using this matrix -- should we update our versions a bit?

```
        strategy:
            matrix:
                include:
                    - php-version: "7.4"
                      drupal: "^8.9"
                      fixture: "d8"
                    - php-version: "7.4"
                      drupal: "^9.2"
                      fixture: "d9"
                    - php-version: "8.0"
                      drupal: "^9.2"
                      fixture: "d9"
```

config/drupal-8/drupal-8-all-deprecations.php
```diff
@@ -27,7 +27,7 @@
     // the following rule adds unneeded annotations.
     $rectorConfig->services()->remove(AddDoesNotPerformAssertionToNonAssertingTestRector::class);
 
-    $rectorConfig->bootstrapFiles([
+    $rectorConfig->autoloadPaths([
         __DIR__ . '/../drupal-phpunit-bootstrap-file.php'
```
mglaman, line 31: We're not passing in paths here; we need a bootstrap file that executes PHP code. The autoload paths scan directories instead. Was there a reason?

mglaman (changes requested):
See https://github.com/palantirnet/drupal-rector/pull/220 for fixes. Required keeping the `bootstrapFiles` but adjusting for Rector's changes when including those bootstrap files. Along with a change in exit code reporting for dry runs.

PR #221, opened 2023-03-28
Identify as rector-extension and default config

mglaman:
## Description
Add `config/config.php` to add default configurations that are required and improve end-user experience.

This will also enable adding services to the service container.

## To Test
Functional run, should provide no changes

## Drupal.org issue
TODO: Needs one


mglaman:
This cannot work since `rector/extension-installer` is part of `rector/rector-src` and the generated config is committed into the Phar

PR #222, opened 2023-03-28
Support providing backward compatibility for Rector rules

mglaman:
## Description
Adds AbstractDrupalCoreRector base class to enable backward compatibility. It also skips Rector rules if the current Drupal version is less than the introduced change.

## To Test
Functional test

## Drupal.org issue
https://www.drupal.org/project/rector/issues/3350886

### DIFF
--- a/src/Rector/AbstractDrupalCoreRector.php
+++ b/src/Rector/AbstractDrupalCoreRector.php
@@ -1,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DrupalRector\Rector;
+
+use DrupalRector\Contract\DrupalCoreRectorInterface;
+use PhpParser\Node;
+use Rector\Core\Rector\AbstractRector;
+
+abstract class AbstractDrupalCoreRector extends AbstractRector implements DrupalCoreRectorInterface
+{
+
+    public function refactor(Node $node)
+    {
+        if (version_compare(\Drupal::VERSION, $this->getVersion(), '<')) {
+            return null;
+        }
+        $result = $this->doRefactor($node);
+        if ($result === null) {
+            return $result;
+        }
+        // TODO: add BC compatibility layer here
+        return $result;
+    }
+
+    /**
+     * Process Node of matched type
+     * @return Node|Node[]|null
+     */
+    abstract protected function doRefactor(Node $node);
+
+}
--- a/src/Rector/AbstractDrupalCoreRector.php
+++ b/src/Rector/AbstractDrupalCoreRector.php
@@ -7,9 +7,22 @@ namespace DrupalRector\Rector;
 use DrupalRector\Contract\DrupalCoreRectorInterface;
 use PhpParser\Node;
 use Rector\Core\Rector\AbstractRector;
+use Rector\PostRector\Collector\NodesToAddCollector;
 
 abstract class AbstractDrupalCoreRector extends AbstractRector implements DrupalCoreRectorInterface
 {
+
+
+    /**
+     * @readonly
+     * @var \Rector\PostRector\Collector\NodesToAddCollector
+     */
+    private $nodesToAddCollector;
+
+    public function __construct(NodesToAddCollector $nodesToAddCollector)
+    {
+        $this->nodesToAddCollector = $nodesToAddCollector;
+    }
 
     public function refactor(Node $node)
     {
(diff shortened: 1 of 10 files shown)


mglaman:
When adding the `user_password` fixture for Drupal 8 we get the expected failure:

```
diff -rubB fixtures/d8/rector_examples_updated/user_password.php /home/runner/drupal/web/modules/custom/rector_examples/user_password.php
--- fixtures/d8/rector_examples_updated/user_password.php	2023-03-28 16:31:32.[9](https://github.com/palantirnet/drupal-rector/actions/runs/4545412070/jobs/8012751417?pr=222#step:10:10)33439759 +0000
+++ /home/runner/drupal/web/modules/custom/rector_examples/user_password.php	2023-03-28 16:32:09.605251754 +0000
@@ -3,8 +3,8 @@
 // should not be changed, deprecated in 9.1.0
 function user_functions() {
     // user_password().
-    $password = user_password();
-    $other_password = user_password(8);
+    $password = \Drupal::service('password_generator')->generate();
+    $other_password = \Drupal::service('password_generator')->generate(8);
     $password_length = [12](https://github.com/palantirnet/drupal-rector/actions/runs/4545412070/jobs/8012751417?pr=222#step:10:13);
-    $last_password = user_password($password_length);
+    $last_password = \Drupal::service('password_generator')->generate($password_length);
 }
```

This change was introduced in 9.1.0 and cannot be fixed on a 8.9 code base

src/Rector/AbstractDrupalCoreRector.php
```diff
@@ -0,0 +1,60 @@
...
+                '>=',
+            ]
+        );
+        $if = new Node\Stmt\If_($versionCompare, [
+            'stmts' => [new Node\Stmt\Expression($result)],
+            'else' => new Node\Stmt\Else_([
+                new Node\Stmt\Expression($node)
+            ]),
+        ]);
+        $this->nodesToAddCollector->addNodeBeforeNode($if, $node);
```
mglaman, line 50: This doesn't work as the `$result` and `$node` are just the replaced parts but ignore the fact it's part of an assignment operation or other context. I think we really do need a PostRector.

src/Rector/AbstractDrupalCoreRector.php
```diff
@@ -0,0 +1,60 @@
...
+
+    public function refactor(Node $node)
+    {
+        if (version_compare(\Drupal::VERSION, $this->getVersion(), '<')) {
+            return null;
+        }
+        $result = $this->doRefactor($node);
+        if ($result === null) {
+            return $result;
+        }
```
mglaman, line 35: We should move this into a compiler pass instead that unregister rules that do not apply to the current version of Drupal core.

mglaman:
Having to implement `PostRector` and possibly a compiler pass to remove unused rules will require additional default config. And in #221 it was proven to be difficult, since we cannot rely on the auto-config for extensions as this is not bundled in Rector

bbrala:
I was thinking on this today, and i think the way to make this managable is to always work on the `Stmt` level. Then replacement and addition is less hard. Although, im not sure if we for example target a method call, if we can replace that method call with a static method call. This would mean writing the below code for multiple types of nodes, but that might be possible.

If we target stmts we can probable do something like this:

1. Find the code on `Stmt` level.
2. Deep clone the old node.
3. Refactor the node.
4. Check if `Stmt` was a `Stmt\Assign` (so we know if we need to return, and if we need to just call `Stmt`, or need to assign the refactored `Stmt` to the var)
5. Create the code needed with `DeprecationHelper::backwardsCompatibleCall(\Drupal::VERSION, $this->getVersion(), [anonimous function with cloned stmt's], [anonimous function with the refactored stmt])`. Depending on the source signature (`Stmt\Assign`, `Stmt\Return_` or just a plain statment), we generate slightly sifferent code.
6. Put the new deprecated call in place of the original node depending on the type above.
7. Profit?

This probably also needs a node visitor that keeps track if the nodes are children of `DeprecationHelper::backwardsCompatibleCall`. This way we can skip those if we need to.

There is also the issue of removing the code, but i think that might be easier since you can look for `DeprecationHelper::backwardsCompatibleCall` where the first `$arg` resolves to an Drupal version below what you want to support.

Now how you make sure you can configure the version you are targetting is a little harder perhaps. I could imagine a ruleset for 9_10 or 10_11, which is pretty much a clone of the higher one, but with the rules 'configured' to target a version. Perhaps if you taget 9_10 then the version would be 9? Which then makes sure there is also a codepath for the old function. If you target 10 (as you might in a Drupal 10 run), you could then just do the default refactor.

Think that is pretty much what would be needed.

bbrala:
Did a POC here with green unit test

https://github.com/palantirnet/drupal-rector/pull/233

agentrickard:
I assume this is now outdated @mglaman @bbrala ?

bbrala:
@mglaman I made a PoC based on this using the class provided by core in #250 

This also includes an "undo" rector that removed the BC calls.

bbrala:
Superceeded by #250

PR #226, opened 2023-07-07
WIP: Fix phpstan issues

agentrickard:
## Description

## To Test

## Drupal.org issue


agentrickard:
Remaining changes:

* https://github.com/search?q=repo%3Arectorphp%2Frector-src+PARENT_NODE&type=commits
* https://github.com/search?q=repo%3Arectorphp%2Frector-src+NEXT_NODE&type=commits
* https://github.com/search?q=repo%3Arectorphp%2Frector-src+findParentType()&type=commits

bbrala:
Some of these will be... interesting. 

Anyways; this might be relevant: https://phpstan.org/blog/preprocessing-ast-for-custom-rules

agentrickard:
Some notes

- I added the neon rule suggested by the phpstan link
- The memory leak concerns seem trivial for our test needs -- we are also on PHPSTAN 1.0 -- so we can handle that in a followup.
- The `PhpParser` library is still using `parent` and `next` keys, so I just hardcoded those.

agentrickard:
I simply don't follow how to replace these calls:

https://github.com/search?q=repo%3Arectorphp%2Frector-src+findParentType()&type=commits

bbrala:
For LinkGeneratorTraitLRector:

```
    public function refactor(Node $node): ?Node
    {
        /** @var Node\Expr\MethodCall $node */
          if ($this->getName($node->name) === 'l') {
              $class = $this->betterNodeFinder->findParentType($node, Node\Stmt\Class_::class);

            // Check if class has LinkGeneratorTrait.
            if ($this->checkClassTypeHasTrait($class, 'Drupal\Core\Routing\LinkGeneratorTrait')) {
              $this->addDrupalRectorComment($node, 'Please manually remove the `use LinkGeneratorTrait;` statement from this class.');
```

The findParentType probably needs to use reflection on the method, then ask reflection the class and from there continue to find the traits. 

Unfortunately there are no tests, and no example. I think this might be because it needed a stub.

bbrala:
Same logic for `DrupalSetMessageRector` i think

bbrala:
`drupal_set_message` it actually tested in the examples, so that might be easier. 

Seems logic is actually a little different, it might need to change de Node\Expr to make sure you can get the class the expression lives in.  Not completely sure without actually debugging the output :P

agentrickard:
I took this as far as I could today. Here's the error on phpunit:

```
5) DrupalRector\Tests\Rector\Deprecation\DrupalSetMessageRector\DrupalSetMessageRectorTest::test with data set #4
assert($classReflection !== null)

/Users/rickard/Sites/drupal-rector/src/Utility/GetDeclaringSourceTrait.php:29
/Users/rickard/Sites/drupal-rector/src/Rector/Deprecation/DrupalSetMessageRector.php:81
/Users/rickard/Sites/drupal-rector/vendor/rector/rector-src/src/Rector/AbstractRector.php:201
```

This likely means that we are passing the wrong var to the method:

```
protected function getDeclaringSource(Node\Expr $expr): ?string
```

bbrala:
Ok, comments thing broke because of this: https://github.com/rectorphp/rector/releases/tag/0.17.3

0.17.2 is not broken. (perhaps its time for some matrix tests on last x releases or something xD)

And i think because of this: https://github.com/rectorphp/rector-src/pull/4463

Now it cannot resolve the parent we change in the comment. So basically, we wither need a personal base which loads that connector if possible, or we have to change the one using comments to always require the node that is worked on to be the Stmt. Then we can add comments.

So i guess, choose your poison ;) 

Thats all the help i can give right now.

mglaman:
One thing I discovered: the current way we attach comments, php-parser ends up treating them as `$origComments` and now the pretty printer breaks since there is no original start position or line.

```
                $comments = $arrItem->getComments();
                $origComments = $origArrItem->getComments();
```

The comments end up on `$origArrItem` when they should be attached our replacement.

mglaman:
Adding context
* https://github.com/nikic/PHP-Parser/issues/791
* https://github.com/rectorphp/rector/discussions/7489

mglaman:
Found more in php-parser PrettyPrinterAbstract:

```php
                if ($comments !== $origComments) {
                    if ($comments) {
                        $result .= $this->pComments($comments) . $this->nl;
                    }
```

The current way comments are added, as previously stated have them treated as _original_ comments. That means `$comments` is an empty array, so the printer acts as if comments were removed from the node.

mglaman:
My thought was to:

* Have `addDrupalRectorComment` set an attribute on the node to add a comment, the attribute value is the comment.
* Have a PostRector which post-processed files to check each `Node\Stmt\Expression` if a child node has the attribute set and apply comments.

We cannot register PostRectors. They are hardcoded in `PostFileProcessor`.

I also tried changing a rule to listen for `Node\Stmt\Expression::class` and verify the expression type (like MethodCall.) Then we'd need to refactor how `addDrupalRectorComment` works or its callers.

mglaman:
Here's an example with checking against the root rexpression: https://github.com/palantirnet/drupal-rector/compare/stan-fails...mglaman:drupal-rector:parent-traverse-fix?expand=1

mglaman:
I wouldn't bump the PHPUnit ruleset to 10 in this PR, so it's more clear on the changes made. Was that required for compatibility?

bbrala:
I'm not sure if this is a good idea. I went in with a fresh head and checked how we would do comments this way. The one failing test is on:

```
$xpath = $this->buildXPathQuery('//select[@name=:name]', [':name' => $name]);
```

Should we rewrite this to an expression the rector rule would need to:

1. Check all expression
2. See if its a Assign
3. If its an assign, check into the assing if there is a method call
4. If its a method call to the magic.

This feels like it will be a LOT of work for such a rule. ;( And there as so many edgecases then. What if its a double assign. Or a assign in some sort of tertiatiari operator. Or on of the other 1000 ways to do this. 

This really feels like its a bad idea, which will complicate things a LOT. Sometimes its easy to rewrite to a statement check, but there will be a lot of missed calls if we need to do it this way. 

It will complicate the code a lot, it will make our rules slower, it will make it less reliable. All because we want to be able to add comments to give extra context on what we might not have fixed.

Another avenue to perhaps explore is using console output for those messages, but I don't think this is really a way we should go.

So basically, i vote to remove the comments. Nice work mglaman, but i feel we are stuck.

mglaman:
We can also add a conflict to the version of Rector so that the package is usable again while this is discussed.

bbrala:
Wooo! Getting there <3

bbrala:
I think we should close this in favor of #238 which is a cleaned up version with also less changes. :)

PR #240, opened 2023-08-18
Upgrade to Rector 0.18.x

bbrala:
Switch to Laravel container means using the container instead of `rectorConfig->parameters()->set`. 

This means that people need to copy rector.php again unfortunately. The container used has changed to Laravel, there is no containerparameters in the Laravel container, so we need to use what Rector switched to.

## Description
Change how enabling comments is done.

## To Test

## Drupal.org issue


bbrala:
Ok, skipping classes needs some work.

bbrala:
After some fighting with rector in a few regards, all if finally green.

Biggest changes:
**New container (`$rectorConfig`) means using a service instead of a container parameter for comments.**
```php
    $rectorConfig->singleton(AddCommentService::class, function() {
        return new AddCommentService(true);
    });
```
This means rector.php needs copying again. Yay ;(

**`TraitsByClassHelperTrait` and `FindByParentTypeTrait` have been refactored away by using a few lines of code.**
```php
$scope = $node->getAttribute(AttributeKey::SCOPE);
$classReflection = $scope->getClassReflection();
if (!is_null($classReflection) && $classReflection->hasTraitUse('Drupal\Core\Routing\LinkGeneratorTrait')) {
 ...
```

**A few rectors no longer needed to ConfigurableRectors since the commenting is now a service**

So it feels like an improvement, and we are back up to head against rector.

rector.php
```diff
@@ -32,5 +31,8 @@
     $rectorConfig->fileExtensions(['php', 'module', 'theme', 'install', 'profile', 'inc', 'engine']);
     $rectorConfig->importNames(true, false);
     $rectorConfig->importShortClasses(false);
-    $parameters->set('drupal_rector_notices_as_comments', true);
+
+    $rectorConfig->singleton(AddCommentService::class, function() {
+        return new AddCommentService(true);
+    });
 };
```
mglaman: Can we add this in the root level of 8.0 or 9.0 config list to prevent a re-copy?

bbrala: We could probably, but we need recopy for removed ->arguments method anyways. 

Also i do want to be able to disable it in update bot. But that might still work, needs testing though.

mglaman: if that's the way it is, so be it.

Configurations with `parameters = $rectorConfig->parameters();` will return an error I'm guessing?

I really wish there was a way to make this easier.

src/Services/AddCommentService.php
```diff
@@ -1,29 +1,24 @@
...
-     * @return void
+     * @param bool $noticesAsComments
      */
-    protected function configureNoticesAsComments(array &$configuration): void
-    {
-        $this->noticesAsComments = $configuration['drupal_rector_notices_as_comments'] ?? false;
-        unset($configuration['drupal_rector_notices_as_comments']);
+    public function __construct(bool $noticesAsComments = true) {
+        $this->noticesAsComments = $noticesAsComments;
     }
```
mglaman, line 22: We never configure it differently, should we just remove the argument?

rector.php
```diff
@@ -32,5 +31,8 @@
     $rectorConfig->fileExtensions(['php', 'module', 'theme', 'install', 'profile', 'inc', 'engine']);
     $rectorConfig->importNames(true, false);
     $rectorConfig->importShortClasses(false);
-    $parameters->set('drupal_rector_notices_as_comments', true);
+
+    $rectorConfig->singleton(AddCommentService::class, function() {
+        return new AddCommentService(true);
```
mglaman: Should we remove the argument if it's always `true`

src/Services/AddCommentService.php
bbrala: It needs disabling in update bot. At least that is how it always worked.

rector.php
bbrala: Yeah it errors, doesnt exist anymore.

What if we include comment service somewhere else? Perhaps we can make it part of the rules? Not sure if rulesets might have access to the container... Im not sure what our other options are.

rector.php
bbrala: As i said before, i'd like to disable it is possible. But that could also perhaps be an empty service so we dont need the config normally?

rector.php
bbrala: Hmm, adding the service in the ruleset might actually work... Then i can still disable it in the rector.php by setting the concrete implementation and would mean we do not need to do this in the future? Let me try that

bbrala:
I could actually relocate the AddCommentService to the rulesets, which means, less noise in the rector.php. Think that is a decent solution since it hides away the service from the developers.

Also tested if i could disable the service in the rector.php, and i can yay.

agentrickard:
@bbrala This tests fine on GitHub and on local. So it seems go to go.

Is this foundational for the next (Drupal 10.1) release?

bbrala:
Nipe, not a base for the ret. Just an upgrade we need to do. See Slack for what I think was a reasonable order.

agentrickard (approved):

PR #241, opened 2023-08-21
StaticArgumentRenameBase and DrupalServiceRenameBase are now covered with StaticArgumentRenameRector

bbrala:
## Description
Migrate to a configured rule for StaticArgumentRenameRector and all child classes.

## Drupal.org issue


src/Rector/ValueObject/StaticArgumentRenameConfiguration.php
```diff
@@ -0,0 +1,38 @@
...
+
+    protected string $methodName;
+
+    protected string $fullyQualifiedClassName;
+
+    protected string $newArgument;
+
+    protected string $oldArgument;
+
+    public function __construct(string $oldArgument, string $newArgument, string $fullyQualifiedClassName = 'Drupal', string $methodName = 'service') {
```
mglaman: Should this default to `\Drupal::service`? I guess maybe that's what "static" refers to? Should it be `StaticServiceArgument...` to be more clear?

mglaman (commented):
Seems like this should be more pointed at service name replacement

src/Rector/ValueObject/StaticArgumentRenameConfiguration.php
bbrala: When built there was a generic base class for the drupal service and one above that.  We could just make this about Drupal service and remove the  premature abstraction.

bbrala: Renames class and arguments to something more sane.

src/Rector/Deprecation/DrupalServiceRenameRector.php
```diff
@@ -0,0 +1,80 @@
...
+        ];
+    }
+
+    public function refactor(Node $node) {
+        if ($node instanceof Node\Expr\StaticCall) {
+            foreach ($this->staticArgumentRenameConfigs as $configuration) {
+                if ($this->getName($node->name) === 'service' && (string) $node->class === 'Drupal') {
+                    if (count($node->args) === 1) {
+                        /* @var Node\Arg $argument */
+                        $argument = $node->args[0];
```
mglaman, line 44: 🤔 I wonder if we can also detect usages of `ContainerInterface::get` here?

We'd need PHPStan's scope, probably. So if `$node->name === 'get'` but only if the class/object is a container object. Maybe as a follow up ticket to this PR so we can land the refactor? Makes it more resilient for any future changes like this.

bbrala: Hmm, we might actually be able to do that yeah. This would also enable refactoring non-global static calls on this which could be great.

I agree on a follow up though, or perhaps somehting we can explore when we start writing rules again.

bbrala: https://www.drupal.org/project/rector/issues/3382531

mglaman: A follow up ticket is good enough. We're not losing functionality, just cleaning up here.

mglaman (approved):

PR #242, opened 2023-08-21
Create FunctionToServiceRector to replace all function to service call deprecations

bbrala:
## Description
Replaces the following rectors with a configured rule.

DrupalRealpathRector.php
DrupalRenderRector.php
DrupalRenderRootRector.php
EntityGetDisplayRector.php
EntityGetFormDisplayRector.php
FileCopyRector.php
FileDirectoryTempRector.php
FileMoveRector.php
FilePrepareDirectoryRector.php
FileSaveDataRector.php
FileScanDirectoryRector.php
FileUnmanagedSaveDataRector.php
FileUriTargetRector.php
FormatDateRector.php
RenderRector.php

## To Test

## Drupal.org issue


src/Rector/Deprecation/FunctionToServiceRector.php
```diff
@@ -0,0 +1,289 @@
...
+    /**
+     * @inheritdoc
+     */
+    public function refactor(Node $node): ?Node
+    {
+        foreach ($this->functionToServiceConfigs as $configuration) {
+            /** @var Node\Expr\FuncCall $node */
+            if ($this->getName($node->name) === $configuration->getDeprecatedFunctionName()) {
+                // This creates a service call like `\Drupal::service('file_system').
+                // TODO use dependency injection.
```
mglaman: I don't know if this `TODO` is a copy over, but I am assuming it'd be near impossible to support dependency injection – is it a class that is a service? Then we have to update `services.yml`. Does it implement `ContainerInjectionInterface`? Then we have to modify `create`.

I think we can remove it because we'll never do it.

mglaman (approved):
Approved, just one minor nit to consider. But it's fine.

src/Rector/Deprecation/FunctionToServiceRector.php
bbrala: Yeah it's a copy. But sure, can remove it while we're at it.

PR #244, opened 2023-08-22
New ExtensionPathRector to replace ExtensionPathBase

bbrala:
## Description
New ExtensionPathRector to replace ExtensionPathBase

## To Test

## Drupal.org issue


bbrala:
Please review, the merge conflict is in the use statements, which will move when earlier PR's are merged. These do not impact the change really. They are just annoying to deal with when doing multiple PR's.

src/Rector/Deprecation/ExtensionPathRector.php
```diff
@@ -0,0 +1,140 @@
...
+                return $node;
+            }
+            [$extensionType, $extensionName] = $args;
+
+            $extensionTypeValueType = $this->nodeTypeResolver->getType($extensionType->value);
+            if ($extensionTypeValueType->getConstantStrings()) {
+                $extensionType = $extensionTypeValueType->getValue();
+            }
+
+            if (in_array($extensionType, [
```
mglaman, line 69: Is it possible `$extensionType` is not a string here? So `!is_string` in case the constant string value couldn't be detected, such as PHPStan considering it as a union (I don't know how or why.)

bbrala: I don't know, i've not audited the code in the Rectors. There might be an edge case imaginable, although i haven't really seen any feedback on this change in any of the regular channels.

When searching gitlab i see this comment nowhere, but since updatebot doesnt do comments this might not mean much.

mglaman: We must not be using PHPStan, because it would/should error on types here.

bbrala: Why not? Needle can be mixed and we check typesafe. So why error? We handle the non-string value "perfectly".

PHPStan is ran on this code.

mglaman: We're at PHPStan level 1

```
parameters:
	level: 1
```

We can merge as is. It'd probably never error, but it could. We can fix later once PHPStan is bumped up

mglaman (approved):
Approving. It works. There is a possible edge case bug, but let's reduce the total number lines of code in this package and then kick up PHPStan level to find these things automatically.

src/Rector/Deprecation/ExtensionPathRector.php
bbrala: Ah, oops. Yeah that is not very helpfull then :x

PR #245, opened 2023-08-22
Remove single use FunctionToImmutableConfigBase

bbrala:
## Description
This base class is used in one concrete implementation. Lets remove it until we might need it.

## To Test

## Drupal.org issue


src/Rector/Deprecation/FileDefaultSchemeRector.php
```diff
@@ -11,13 +12,45 @@
...
+final class FileDefaultSchemeRector extends AbstractRector
 {
-    protected $deprecatedFunctionName = 'file_default_scheme';
+    protected string $deprecatedFunctionName = 'file_default_scheme';
 
-    protected $configObject = 'system.file';
+    protected string $configObject = 'system.file';
 
-    protected $configName = 'default_scheme';
+    protected string $configName = 'default_scheme';
```
mglaman, line 21: Do we need these properties anymore, then? Or do we want to keep it as a concept for refactoring into a configurable rule, later?

bbrala: Currently its a laser focussed rector. So keep them in the class, exactly because this makes recatoring to confgurable a lot easier should we choose to later. Also it makes deciding if this logic is useable a little easier since you know what the parameters are.

mglaman (approved):

PR #247, opened 2023-08-22
MethodToMethodWithCheckRector to replace methods with a certainty check

bbrala:
## Description
MethodToMethodWithCheckRector to replace methods with a certainty check.

## To Test

## Drupal.org issue


bbrala:
Don't understand why this is failing for toUrl. Suddenly the code is sure of the super class? What gives! :/

https://github.com/palantirnet/drupal-rector/pull/247/files#diff-390fa78faf8fb7c47bc612a2737896255adca35cc53fdf21f536b138ed3f2338R110

src/Rector/Deprecation/MethodToMethodWithCheckRector.php
```diff
@@ -0,0 +1,187 @@
...
+        if ($isSuperOf->yes()) {
+            $node->name = new Node\Identifier($configuration->getMethodName());
+            return $node;
+        }
+
+        if ($isSuperOf->maybe()) {
+            $node_var = $node->var->name;
+
+            if ($node->var instanceof Node\Expr\Variable) {
+                $node_var = "$$node_var";
```
mglaman, line 114: Do you know if something needs to be escaped here? Because we're doing that weird PHP thing with double `$$`.

Would this be `$node_var = "${$node_var}";`

bbrala: It is kinda weird to do something like this while we should have the power of rector and phpstan. 

Perhaps I should just try and add the failing case to a unit test, if that also fails debugging is a little easier.

bbrala: Well, it's only for th printing, and it seems to do nothing special.

bbrala:
Well, its green. Seems the d8 code is understood better for some reason.

mglaman (approved):

mglaman (dismissed):
phpstan errors :/

bbrala:
ill fix those i guess, these are not that hard to fix mostly.

bbrala:
Made code more readable by phpstan by making it less readable for humans. Thanks phpstan.

agentrickard (approved):

PR #250, opened 2023-09-04
Support for backwards compatible rector rules and version scoping of rules

bbrala:
## Description
Adds AbstractDrupalCoreRector base class to enable backward compatibility. It also skips Rector rules if the current Drupal version is less than the introduced change.

Based on the work  by @mglaman on #222 and refactored since we now have `DeprecationHelper` in core.

This change introduced a number of new things:

### Configuration interface for BC enabled rules
`DrupalRector\Contract\VersionedConfigurationInterface`:  If an configuration implements this interface the developer can supply a version for the configured rule. With this drupal-rector can identify the following things:
1. What version does this rule apply to
2. What version needs a BC fix
3. What version just needs the normal fix.

### Concrete configuration for rules with no own configuration
`DrupalRector\Rector\ValueObject\DrupalIntroducedVersionConfiguration`: A configuration valueobject that only contains a version. If a rector does not need any configuration itself, the developer can use this valueobject to supply version information about the rector.

### Base class for BC rule rectors
`AbstractDrupalCoreRector`: This class can be used as a base for BC fixes. It will check the Drupal version, the configuration object for applicable version and either skip the rule, run in in BC mode, or just run it normally. 

#### Supported node types
Currently the only type of node it applies to is `CallLike`. When new rules arrise that might need to apply to other type of nodes. But lets add those as they come up.

```php
// Create a backwards compatible call if the node is a call-like expression.
if ($node instanceof Node\Expr\CallLike && $result instanceof Node\Expr\CallLike) {
  return $this->createBcCallOnCallLike($node, $result, $configuration->getIntroducedVersion());
}
```

CallLike is a one of the following type of calls:
```php
PhpParser\Node\Expr\FuncCall::class
PhpParser\Node\Expr\MethodCall::class
PhpParser\Node\Expr\New_::class
PhpParser\Node\Expr\NullsafeMethodCall::class
PhpParser\Node\Expr\StaticCall::class
```

#### Future optimization with SkipVoters
In the future we might want to add a SkipVoterInterface so we can determine that a rule does not apply earlier in the process. BUT this means if old code is introduced into a project at a later time, that will not be removed. So not sure we even want that.

#### Rector to remove BC code paths
`DrupalRector\Rector\Deprecation\RemoveDeprecationHelperRector`: This rector will remove all BC calls using `DeprecationHelper` from the code if the Drupal version used is higher than the version check in the code. For now it is only ran in a UnitTest, but as we get closed to Drupal 11, we might want to enable this.

## To Test
- Unit tests
- Functional tests

## Drupal.org issue
https://www.drupal.org/project/rector/issues/3350886


bbrala:
Problaby missing some cases, what if the BC layer is more than one call? What if the match is a return?  What if child nodes are modified in the matched node, does a simple clone break then because its not a deep clone?

But this also gives a base to create unit tests that check those kind of deprecations.

bbrala:
One issue this made me realize, there are also deprecations in for example 9.5 that will only be removed in 11. Which is pretty annoying, since that means 2 major versions. ;(

We might need to supply that at some point, but since we are not doing 9.5, me might be OK for 10 -> 11.

bbrala:
If we agree on this pattern, we can rebase this onto #252 and remove the `user_password` example and add the pattern to the Drupal 10.1 deprecation.

This will mean though that "version" must be something that is registered in the configuration of that specific rule, since it is used for an 8.x deprecation and a later one.

phpstan-baseline.neon
```diff
@@ -35,6 +35,11 @@ parameters:
 			count: 1
 			path: config/drupal-phpunit-bootstrap-file.php
 
+		-
+			message: "#^Else branch is unreachable because ternary operator condition is always true\\.$#"
+			count: 1
+			path: src/Rector/AbstractDrupalCoreRector.php
+
```
mglaman: If we add the following to `phpstan.neon` this won't be reported:

```
	dynamicConstantNames:
		- Drupal::VERSION
```

See https://github.com/mglaman/phpstan-drupal/blob/main/extension.neon#L14-L15

src/Rector/AbstractDrupalCoreRector.php
```diff
@@ -0,0 +1,52 @@
...
+use DrupalRector\Contract\DrupalCoreRectorInterface;
+use PhpParser\Node;
+use PhpParser\Node\Expr\ArrowFunction;
+use Rector\Core\Rector\AbstractRector;
+
+abstract class AbstractDrupalCoreRector extends AbstractRector implements DrupalCoreRectorInterface
+{
+    public function refactor(Node $node)
+    {
+        $drupalVersion = substr(\Drupal::VERSION, -4) === '-dev' ? str_replace(['.x-dev', '-dev'], '.0', \Drupal::VERSION) : \Drupal::VERSION;
```
mglaman: There is no penalty to run `str_replace` on a non-matching string, so we can simplify this logic and make it easier to read

mglaman (commented):
Some initial thoughts, brain couldn't upload easily right now.

phpstan-baseline.neon
bbrala: Fixed

src/Rector/AbstractDrupalCoreRector.php
bbrala: Fixed

src/Rector/AbstractDrupalCoreRector.php
```diff
@@ -0,0 +1,52 @@
...
+use PhpParser\Node;
+use PhpParser\Node\Expr\ArrowFunction;
+use Rector\Core\Rector\AbstractRector;
+
+abstract class AbstractDrupalCoreRector extends AbstractRector implements DrupalCoreRectorInterface
+{
+    public function refactor(Node $node)
+    {
+        $drupalVersion = str_replace(['.x-dev', '-dev'], '.0', \Drupal::VERSION);
+        if (version_compare($drupalVersion, $this->getVersion(), '<')) {
```
mglaman: My brain says: shouldn't `.0` be `.9999` but I don't know. I guess its best to assume `.0` because we don't know where or what the `dev` is at. It could be an old HEAD. For GitLab CI/DrupalCI `drupal/core` isn't always a tagged release, will this break project update bot because it'll skip a lot of things?

src/Rector/Deprecation/RemoveDeprecationHelperRector.php
```diff
@@ -0,0 +1,97 @@
...
+$pass = user_password();
+$shorter_pass = user_password(8);
+CODE_BEFORE
+                ,
+                <<<'CODE_AFTER'
+$pass = \Drupal::service('password_generator')->generate();
+$shorter_pass = \Drupal::service('password_generator')->generate(8);
+CODE_AFTER
+            )
+        ]);
```
mglaman: I don't think this matches. It's a Rector to Rector our replacements of `backwardsCompatibleCall`?

mglaman: Or is this a Rector to revert BC calls when they're no longer needed?

src/Rector/AbstractDrupalCoreRector.php
bbrala: Core policy defines that depreactions can only be added in minors, see https://www.drupal.org/about/core/policies/core-change-policies/allowed-changes#minor

So lets see:
Depraction introduced in 10.1.0 should evalutate to false on anything on or after that version. This means that lower than 10.1.0 is correct. Since if you are running on 10.1.x you will evaluate to false.

src/Rector/Deprecation/RemoveDeprecationHelperRector.php
bbrala: Yes this rector is to remove BC calls when you no longer need those. What if we are at Drupal 11 -> 12 cycle, then we need to probably remove Drupal 10 -> 11 BC calls at some point.

Also if a new major version of your module is released this could help you remove all the BC calls easily if you want to drop a major version

bbrala: But you are also right. This example is not correct and need to be updated with a depracatedhelper call to a non-deprecated helper call. Oops.

bbrala: I wil fix this while rebasing and using a true Drupal 10 deprecataion

src/Rector/AbstractDrupalCoreRector.php
mglaman: Ah, yeah, you're right :)

src/Rector/Deprecation/RemoveDeprecationHelperRector.php
bbrala: Resolved

README.md
```diff
@@ -12,6 +12,11 @@ Automate fixing deprecated Drupal code.
 
 ### Release notes
 
+* The 0.18.0 and higher releases of drupal-rector will include Rector 0.18+. The upgrade path should be as simple as re-copying the configuration file. `cp vendor/palantirnet/drupal-rector/rector.php`.
+
+> [!WARNING]
+> Drupal rector does not support Drupal 10.0 specifically because this will break support for backwards compatible fixes for which the helpers were introduced in Drupal 10.1. Older and newer versions of Drupal are supported.
```
mglaman: This confused me, what specifically was changed here? Why can't we just fix them as non-backward compatible like previous version?

src/Rector/AbstractDrupalCoreRector.php
```diff
@@ -0,0 +1,90 @@
...
+
+            // Check if Drupal version and the introduced version support backward
+            // compatible calls. Although it was introduced in Drupal 10.1 we
+            // also supply these patches for changes introduced in Drupal 10.0.
+            // The reason for this is that will start supplying patches for
+            // Drupal 10 when 10.0 is already out of support. This means that
+            // we will not support running drupal-rector on Drupal 10.0.x.
+            if (version_compare($drupalVersion, '10.1.0', '<') || version_compare($configuration->getIntroducedVersion(), '10.0.0', '<')) {
+                return $result;
+            }
```
mglaman, line 60: Ah I think I get it. We're going to replace 10.0.x code with backwards compatibility introduced in 10.1.x because this will be run using versions of Drupal greater than 10.0.x (in bulk)

Also, ugh. I realize this means modules must require a minimum of 10.1.x if they use it. They cannot do a `^9.5 || ^10 || ^11` bridge. Oh well

src/Rector/Deprecation/FunctionToStaticRector.php
```diff
@@ -17,12 +18,12 @@
...
  * - Static replacement
  */
-class FunctionToStaticRector extends AbstractRector implements ConfigurableRectorInterface
+class FunctionToStaticRector extends AbstractDrupalCoreRector
 {
     /**
      * @var array|FunctionToStaticConfiguration[]
      */
-    private array $configuration;
+    protected array $configuration;
```
mglaman, line 26: Does this still need to be defined if the parent has it? Ah, for typing analysis it does.

tests/src/Rector/Deprecation/DeprecationHelperRemoveRector/fixture/user_password_keep.php.inc
mglaman, line 1: This is the same as `user_password.php.inc`. It says `_keep` so I assumed it should keep the backwards compatible call. But it doesn't

tests/src/Rector/Deprecation/DeprecationHelperRemoveRector/fixture/user_password_keep_simplefqcn.php.inc
mglaman, line 1: What oes this test do? it says `simplefqcn` but I don't  see it doing anything about fully qualified class names.

README.md
bbrala: Yeah we can, but question is if we want that split. Was also thinking about update not. But I'll just run 10.1 and adjust changes to the info.yml accordingly.

 So I will remove this part

src/Rector/AbstractDrupalCoreRector.php
bbrala: Yeah we do unfortunately. So it would be:

^9.5 | ^10.1 | ^11

But that's not for local code perse, I'll fix updatebot to do that though

src/Rector/Deprecation/FunctionToStaticRector.php
bbrala: Yeah it a more specific type indeed.

tests/src/Rector/Deprecation/DeprecationHelperRemoveRector/fixture/user_password_keep.php.inc
bbrala: oops

tests/src/Rector/Deprecation/DeprecationHelperRemoveRector/fixture/user_password_keep_simplefqcn.php.inc
bbrala: hmm, that was an issue at some point. Forgot to clean up

tests/src/Rector/Deprecation/DeprecationHelperRemoveRector/fixture/user_password_keep.php.inc
```diff
@@ -0,0 +1,27 @@
+<?php
+
+use Drupal\Component\Utility\DeprecationHelper;
+
+function simple_example()
+{
+    $password = DeprecationHelper::backwardsCompatibleCall(\Drupal::VERSION, '10.5.0', fn() => user_password(), fn() => \Drupal::service('password_generator')->generate());
```
mglaman, line 7: 👍 so if 10.5.0 is released or when 11.0.0 is released, this test _should_ fail.

mglaman (approved):
I'm going to approve! This looks great. It's a huge improvement to Rector. Before it just fixed anything regardless the version of Drupal.

I guess most rules _still_ do, but now we're able to properly scope rules based on the version of Drupal core! Folks can use drupal-rector without being on the latest minor version of Drupal core!

tests/src/Rector/Deprecation/DeprecationHelperRemoveRector/fixture/user_password_keep.php.inc
bbrala: Hehe technically, but unit tests don't use a real install but a shim for the version. ;)

PR #251, opened 2023-09-05
Refactor Base\FunctionToStatic to configurable rule.

bbrala:
## Description
This one will be needed a lot in Drupal 10. By making it configurable before starting on d10 to d11 we save us a lot of time.

Want to use this one as an example d10 -> d11 rector rules. And perhaps as an example for a configurable rule with version in it.

## To Test

## Drupal.org issue


src/Rector/ValueObject/FunctionToStaticConfiguration.php
```diff
@@ -0,0 +1,35 @@
...
+
+namespace DrupalRector\Rector\ValueObject;
+
+class FunctionToStaticConfiguration {
+
+    protected string $deprecatedFunctionName = 'file_directory_os_temp';
+
+    protected string $className = 'Drupal\Component\FileSystem\FileSystem';
+
+    protected string $methodName = 'getOsTemporaryDirectory';
```
mglaman: Should this defaults exist?

mglaman (dismissed):
The config shouldn't have default values for the properties, should it?

src/Rector/ValueObject/FunctionToStaticConfiguration.php
bbrala: yeah it shouldnt, those leaked... will fix.

bbrala: Removed the default from the confguration.

agentrickard (approved):

PR #252, opened 2023-09-05
Support for Drupal 10 deprecations and testing

bbrala:
## Description
For now only run this in tests until we figure out how to also apply the deprecation helper. We need to decide how we handle the fact that this helper is only implemented in Drupal 10.1, perhaps we should not support Drupal 10.0 and only support Drupal 10.1 and up?

This adds:
1. Deprecation files for Drupal 10
2. Correct setlists for Symfony for 10.0 and 10.0
3. First deprecation (10.1 though): https://www.drupal.org/node/3244583
4. GitHub action configuration for Drupal 10 functional testing.

We need to decide the following:
1. How to handle what rules are ran on the project. We don't want D10 stuff running on D9. We could detect the version, or perhaps have some other way to influence what rector.php is used. (less copying, more configuring by the user maybe?)
2. This does not include the Deprecation helper for now, that is only  Drupal 10.1, which means we need to choose if we even support Drupal 10.0, since that will complicate things a lot. And to include https://packagist.org/packages/drupal/core-utility#10.1.0 is not really realistic, so we are in a bit of a conundrum here.

## To Test
- Unit tests
- Functional tests

## Drupal.org issue
[Provide a link to the issue from https://www.drupal.org/project/rector/issues. If no issue exists, please create one and link to this PR.](https://www.drupal.org/project/rector/issues/3384046)

### DIFF
--- a/config/drupal-10/drupal-10.1-deprecations.php
+++ b/config/drupal-10/drupal-10.1-deprecations.php
@@ -11,9 +11,7 @@ use Rector\Symfony\Set\TwigLevelSetList;
 
 return static function (RectorConfig $rectorConfig): void {
     $rectorConfig->sets([
-        PHPUnitLevelSetList::UP_TO_PHPUNIT_90,
         SymfonyLevelSetList::UP_TO_SYMFONY_63,
-        TwigLevelSetList::UP_TO_TWIG_240,
     ]);
 
     // https://www.drupal.org/node/3244583
--- a/config/drupal-10/drupal-10.1-deprecations.php
+++ b/config/drupal-10/drupal-10.1-deprecations.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 
 use DrupalRector\Rector\Deprecation\FunctionToStaticRector;
 use DrupalRector\Rector\ValueObject\FunctionToStaticConfiguration;
-use DrupalRector\Set\Drupal10SetList;
 use Rector\Config\RectorConfig;
 use Rector\PHPUnit\Set\PHPUnitLevelSetList;
 use Rector\Symfony\Set\SymfonyLevelSetList;
--- a/config/drupal-10/drupal-10.1-deprecations.php
+++ b/config/drupal-10/drupal-10.1-deprecations.php
@@ -1,0 +1,24 @@
+<?php
+
+declare(strict_types=1);
+
+use DrupalRector\Rector\Deprecation\FunctionToStaticRector;
+use DrupalRector\Rector\ValueObject\FunctionToStaticConfiguration;
+use DrupalRector\Set\Drupal10SetList;
+use Rector\Config\RectorConfig;
+use Rector\PHPUnit\Set\PHPUnitLevelSetList;
+use Rector\Symfony\Set\SymfonyLevelSetList;
+use Rector\Symfony\Set\TwigLevelSetList;
+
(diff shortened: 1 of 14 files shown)


config/drupal-10/drupal-10.1-deprecations.php
```diff
@@ -0,0 +1,23 @@
...
+use Rector\PHPUnit\Set\PHPUnitLevelSetList;
+use Rector\Symfony\Set\SymfonyLevelSetList;
+use Rector\Symfony\Set\TwigLevelSetList;
+
+return static function (RectorConfig $rectorConfig): void {
+    $rectorConfig->sets([
+        PHPUnitLevelSetList::UP_TO_PHPUNIT_90,
+        SymfonyLevelSetList::UP_TO_SYMFONY_63,
+        TwigLevelSetList::UP_TO_TWIG_240,
+    ]);
```
mglaman: Do we need these here if in `config/drupal-10/drupal-10.0-deprecations.php`?

config/drupal-8/drupal-8.2-deprecations.php
```diff
@@ -2,9 +2,16 @@
...
 return static function (RectorConfig $rectorConfig): void {
-    $rectorConfig->rule(FileDirectoryOsTempRector::class);
+    // https://www.drupal.org/node/2802569
+    $rectorConfig->ruleWithConfiguration(FunctionToStaticRector::class, [
+        new \DrupalRector\Rector\ValueObject\FunctionToStaticConfiguration(
+            'file_directory_os_temp',
+            'Drupal\Component\FileSystem\FileSystem',
+            'getOsTemporaryDirectory'
+        ),
+    ]);
```
mglaman: This is from the other PR, right?

mglaman (commented):
Looks like this will be simplified once https://github.com/palantirnet/drupal-rector/pull/251 is merged

config/drupal-8/drupal-8.2-deprecations.php
mglaman: https://github.com/palantirnet/drupal-rector/pull/251

config/drupal-10/drupal-10.1-deprecations.php
bbrala: You are right, its not consistent with the other rulesets. Just need the symfony one since that minor got upped.

config/drupal-8/drupal-8.2-deprecations.php
bbrala: Yeah same code, wanted that as the base for this one, since we going to need that a lot and its pretty simple.

Still, we might merge the other one before this anyways. Here we still need to decide how to handle the fact you migght run this on d9. Then the whole 'copy rector.php' is either invalid, or we need to do magic. Think i might prefer as little magic as possible ;p

bbrala: Also for testing we might need specific rector.php's i think, since the new phpunit set adds a comment to the current drupal 9 set.

bbrala: #251 has been merged.

bbrala:
Rebased this on main

mglaman (approved):
Looks good to me!

bbrala:
Yay!

PR #255, opened 2023-09-21
Refactored BC rules with configuration

bbrala:
## Description
Draft of refactor for the whole BC rules to review before


mglaman (commented):
This makes sense to me. The Drupal version needs to be part of configuration. Which means we'll also need a generic configuration that accepts just a version number.

This adds overhead but makes this tool usable for contrib. The Drupal community has decided this should be used for contrib, so let's make our contrib tooling better and life better for maintainers.

PR #257, opened 2023-10-29
WIP: Rule to convert action to attributes

andypost:
## Description

Rule to convert Action annotation to Attribute https://www.drupal.org/node/3395575

To-do https://www.drupal.org/node/3229001

## To Test

using with core and examples module
`vendor/bin/rector --xdebug -c rector.php -n process modules_/examples/modules/action_example`

## Drupal.org issue

https://www.drupal.org/project/rector/issues/3265946


andypost:
```diff
1 file with changes
===================

1) modules_/examples/modules/action_example/src/Plugin/Action/BasicExample.php:1

    ---------- begin diff ----------
@@ @@

 namespace Drupal\action_example\Plugin\Action;

+use Drupal\Core\Action\Attribute\Action;
+use Drupal\Core\StringTranslation\TranslatableMarkup;
 use Drupal\Core\Access\AccessResult;
 use Drupal\Core\Action\ActionBase;
 use Drupal\Core\Messenger\MessengerInterface;
@@ @@

 /**
  * A basic example action that does nothing.
- *
- * @Action(
- *   id = "action_example_basic_action",
- *   label = @Translation("Action Example: A basic example action that does nothing"),
- *   type = "system"
- * )
  */
+#[Action(id: 'action_example_basic_action', label: new TranslatableMarkup('Action Example: A basic example action that does nothing'), type: 'system')]
 class BasicExample extends ActionBase implements ContainerFactoryPluginInterface {

   /**
    ----------- end diff -----------

Applied rules:
 * ActionAnnotationToAttributeRector (https://docs.phpunit.de/en/10.0/annotations.html#ticket)
```

bbrala:
Good start :heart:, thank you for working on this.

I'd probably make it a `ConfigurableRector` for one where you can feed it a configuration valueobject. I'm assuming the input for this rector would always be the tagname, since that should always map 1-on-1 to the attribute (class)? (see [FunctionToStaticValueObject](https://github.com/palantirnet/drupal-rector/blob/main/src/Rector/ValueObject/FunctionToStaticConfiguration.php) that implemets [VersionedConfigurationInterface](https://github.com/palantirnet/drupal-rector/blob/main/src/Contract/VersionedConfigurationInterface.php)) for an example of such an implementation.

Then it could work like a backwards compatible rule, where we pass the version it was introduced in ([as FunctionToStaticRector](https://github.com/palantirnet/drupal-rector/blob/main/src/Rector/Deprecation/FunctionToStaticRector.php)). But it also means it needs new handling in the [AbstractDrupalCoreRector](https://github.com/palantirnet/drupal-rector/blob/main/src/Rector/AbstractDrupalCoreRector.php) since the BC call is not really a call (which can be called using DeprecationHelper), but the fact it does not remove the docblock annotatation. Thinking about that, it might need some refactoring to handle this case of it being a docblock, or we need to handle that in the rector itself in this case. But i need to be more awake to think about that one.

Random thought, what if we make the removal bases on a property of ActionAnnotationToAttributeRector which we can set with a setter? Then we can toggle it easily. By default remove the docblock, but allow `$rector->setDocblockRemoval(false)` to disable it? This could be an extra argument in the `ValueObject` also maybe that forces the removal/keep behaviour?

The end result would be:

1. A ConfigurableRector that can handle refactoring any of the coming annotation changes. As long as we are consistent in the change we just add a new rule with the correct configuration
2. A way to provide the change with or without keeping the original docblock (BC fixes yay).
3. Easy way to convert core (which can be configurable locallly to force removal even if you are running it against 10.x)
4. Profit.

When looking at the core change record i did also notice the following:

```php
/*
* @Block(
 *   id = "test_block_instantiation",
 *   admin_label = @Translation("Display message")
 * )
 */
#[Block(
  id: "test_block_instantiation",
  admin_label: new TranslatableMarkup("Display message")
)]
```

This does mean there is nested tags going on? Do we know what those are? Is that a fixed set?

andypost:
There's issue about formatting as miltiline https://github.com/rectorphp/rector/issues/8239

andypost:
The set keys of annotation is defined by class https://git.drupalcode.org/project/drupal/-/blob/11.x/core/lib/Drupal/Core/Block/Annotation/Block.php

Technically the annotation could be parsed, child nodes could be picked from properties' doc-blocks and mapped
- `@var \Drupal\Core\Annotation\Translation`
- `@var \Drupal\Core\Annotation\ContextDefinition[]`

src/Rector/Deprecation/ActionAnnotationToAttributeRector.php
```diff
@@ -0,0 +1,171 @@
...
+        }
+        return null;
+    }
+    private function createAttribute(array $parsedArgs) : Attribute
+    {
+        $fullyQualified = new FullyQualified(Action::class);
+        $args = [];
+        foreach ($parsedArgs as $value) {
+            if ($value->key === 'label') {
+                $arg = new Node\Expr\New_(new Node\Name(TranslatableMarkup::class), [new Arg(new String_($value->value->values[0]->value->value))]);
```
andypost: it could be a helper method

Berdir:
> There's issue about formatting as miltiline [rectorphp/rector#8239](https://github.com/rectorphp/rector/issues/8239)

I don't think that's related. That's not about generating multiline attributes, it's  about parsing annotations with multiline strings.

andypost:
To apply coding standards, according to https://github.com/rectorphp/rector/blob/main/README.md#how-to-apply-coding-standards

https://github.com/rectorphp/rector-src/blob/main/ecs.php

src/Rector/Deprecation/ActionAnnotationToAttributeRector.php
```diff
@@ -0,0 +1,171 @@
...
+            }
+            else {
+                $arg = new String_($value->value->value);
```
mstrelan: In our internal projects we tend to use a `const PLUGIN_ID = 'foo'` pattern and reference these in our annotations. I've been playing around with a custom annotation to attribute rector rule based on this one and noticed it fails on these constants. The following seems to work, although I'm wondering if there is a better way to detect the use of a constant in the annotation.

```
$arg = new String_($value->value instanceof StringNode ? new String_($value->value->value) : new ConstFetch(new Node\Name($value->value)));
```

config/drupal-10/drupal-10.2-deprecations.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+use DrupalRector\Drupal10\Rector\Deprecation\ActionAnnotationToAttributeRector;
+use Rector\Config\RectorConfig;
+
+return static function (RectorConfig $rectorConfig): void {
+    // @see https://www.drupal.org/node/3395575
+    $rectorConfig->rule(ActionAnnotationToAttributeRector::class);
```
mglaman: Should this specify the introduced version?

src/Drupal10/Rector/Deprecation/ActionAnnotationToAttributeRector.php
```diff
@@ -0,0 +1,270 @@
...
+        }
+
+        return null;
+    }
+
+    /**
+     * @param Class_|ClassMethod                             $node
+     * @param DrupalIntroducedAndRemovalVersionConfiguration $configuration
+     */
+    public function refactorWithConfiguration(Node $node, VersionedConfigurationInterface $configuration): ?Node
```
mglaman: Does this need to be public?

bbrala:
What changed yesterday (AI generated):

1. In the ActionAnnotationToAttributeRector.php file, the convertTranslateAnnotation method had its input parameter type and internal logic modified to work directly with the actual annotation instead of its container, simplifying the conversion process.
2. Another significant modification is made in handling method annotation-to-attribute conversions. To support backward compatibility in refactors, the changes allow annotative refactorings only on Drupal version 10.1 or higher, ignoring lower versions.
3. AbstractDrupalCoreRector.php was refactored to improve the readability of the methods, ensuring return types match the method names. It also checks if the Drupal version supports backward compatibility.
4. The Drupal version in Drupal.php (\Drupal::VERSION) was updated to '10.2.x-dev'.

These changes seem to streamline the annotation handling in Drupal and improve compatibility with earlier versions.

Note:
I do think at some points we'd want a service to get the Drupal version. Then we can actually set the version we run against in the tests if we like, which will make a few things easier.

andypost:
New conversions
- `ImageToolkitOperation` https://git.drupalcode.org/project/drupal/-/commit/328bb474d7
- `ImageToolkit` https://git.drupalcode.org/project/drupal/-/commit/ffaf9871607dbfe42c976da2109b0d04f15faada
- `ImageEffect` https://git.drupalcode.org/project/drupal/-/commit/58ecf6d433

andypost:
Added the rest of conversion https://www.drupal.org/node/3229001

src/Drupal10/Rector/Deprecation/AnnotationToAttributeRector.php
```diff
@@ -0,0 +1,282 @@
...
+            }
+
+            if ($hasAttribute === false) {
+                $stringValue = $valueNode->value->value;
+                $stringValue = '{'.trim($stringValue, '()').'}';
+                $tokenIterator = $this->tokenIteratorFactory->create($stringValue);
+                $data = $this->arrayParser->parseCurlyArray($tokenIterator, $node);
+                $attribute = $this->createAttribute($configuration->getAttributeClass(), $data);
+                $node->attrGroups[] = new AttributeGroup([$attribute]);
+            }
```
mglaman, line 198: Can this do a better check that the annotation matching the ending class name?

Wouldn't this cause `@foo` to become `#Foo` attribute because its a generic string?

src/Drupal10/Rector/Deprecation/ActionAnnotationToAttributeRector.php
bbrala: No, oops. 

`abstract protected function refactorWithConfiguration`

config/drupal-10/drupal-10.2-deprecations.php
bbrala: removed while back, thx

src/Rector/Deprecation/ActionAnnotationToAttributeRector.php
bbrala: See root comment.

bbrala:
Perhaps we could do more with this one by rector itself:

`AnnotationToAttributeMapper $annotationToAttributeMapper`
`$attribute = $this->annotationToAttributeMapper->map($value);`

This class (which allowd extending classes by interface) does smart things with values. It also does stuff like mapping consts and such.

Kinda worth exploring if we can maybe just hook into that for converting some. The code looked very interesting to structure the mapping. Those classes were made for doctrine itself.

src/Drupal10/Rector/Deprecation/AnnotationToAttributeRector.php
bbrala: Do you mean here; you get to this code after checking a tag is present as configured. Then we move on to check if there is already an attribute. If that is present, then we loopt through the tagnames and if there was no attribute matching the relevant attribute class from the configuration we start building the attribute based on the class passed into the rector by the configuration. 

Don't really see how you could break this, unless user error using their own config?

bbrala:
> New conversions
> 
> * `ImageToolkitOperation` https://git.drupalcode.org/project/drupal/-/commit/328bb474d7
> * `ImageToolkit` https://git.drupalcode.org/project/drupal/-/commit/ffaf9871607dbfe42c976da2109b0d04f15faada
> * `ImageEffect` https://git.drupalcode.org/project/drupal/-/commit/58ecf6d433

BTW want to mention, these three would've been automatable with this MR without any problem. All known properties.

src/Drupal10/Rector/Deprecation/AnnotationToAttributeRector.php
mglaman: More so just:

```
$blah = explode('\', $configuration->getAttributeClass());
$className = end($blah);
if ($blah === substr($stringValue, 1)) {
```

Not entirely configured. Just that the tag matches the attribute's class name. I'm going to bet there is a "malformed" phpdoc somewhere with what the phpdoc parser considers generic tags

mglaman (approved):
Approving. The rule isn't configured and this allows using it for Drupal core improvements.

PR #260, opened 2023-11-10
Fix MethodToMethodWithCheckRector by also matching on MethodCall.

bbrala:
## Description
Also match on Node\Expr\MethodCall so we can match deep matches. 

## To Test
- See tests

## Drupal.org issue
https://www.drupal.org/project/rector/issues/3400231

### DIFF
--- a/src/Rector/Deprecation/MethodToMethodWithCheckRector.php
+++ b/src/Rector/Deprecation/MethodToMethodWithCheckRector.php
@@ -76,11 +76,11 @@ class MethodToMethodWithCheckRector extends AbstractRector implements Configurab
                 continue;
             }
 
-            if ($node->expr instanceof Node\Expr\MethodCall && $this->getName($node->expr->name) !== $configuration->getDeprecatedMethodName()) {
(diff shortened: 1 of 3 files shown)


src/Rector/Deprecation/MethodToMethodWithCheckRector.php
```diff
@@ -57,28 +57,41 @@ public function configure(array $configuration): void {
...
             }
 
-            if ($node->expr instanceof Node\Expr\Assign && $node->expr->expr instanceof Node\Expr\MethodCall && $this->getName($node->expr->expr->name) !== $configuration->getDeprecatedMethodName()) {
+            if ($node instanceof Node\Stmt\Expression && $node->expr instanceof Node\Expr\MethodCall && $this->getName($node->expr->name) !== $configuration->getDeprecatedMethodName()) {
                 continue;
             }
 
+            if ($node instanceof Node\Stmt\Expression && $node->expr instanceof Node\Expr\Assign && $node->expr->expr instanceof Node\Expr\MethodCall && $this->getName($node->expr->expr->name) !== $configuration->getDeprecatedMethodName()) {
+                continue;
+            }
```
mglaman, line 85: I feel like this could be optimized to pop out the expressions and find the method call, but that can be done later for this entire rule.

mglaman (approved):

PR #267, opened 2023-11-11
Restructure rules by major and generate rule list (docs)

bbrala:
## Description
Refactored al rules to namespace by major like rector does in its packages. This makes a lot of sense in regard to overview of what rules are actually recent.

Also added composer commands for convenience:

```
composer docs
composer tests
composer phpstan
```

There are a few rectors that are 'generic' and not specific to a single version. Not sure yet if its best to keep ik like that, or make them abstract and extend them in the Major namespace. But this is a good step forward.

Added generation of docs from the ruleset, saved here: https://github.com/palantirnet/drupal-rector/blob/feature/restructure-by-major/docs/rules_overview.md

## To Test
- Tests are green
- use `composer docs` to update the docs.

## Drupal.org issue
Nope


mglaman:
I like that version specific rectors are in their own namespace. Makes it easier to refactor them all into configured rules unless they truly are snowflakes.

Offhand it looks good from mobile. I want to just give it a second look over from desktop later

config/drupal-9/drupal-9.1-deprecations.php
```diff
@@ -3,19 +3,19 @@
...
+use DrupalRector\Drupal9\Rector\Deprecation\AssertNoFieldByNameRector;
+use DrupalRector\Drupal9\Rector\Deprecation\AssertNoUniqueTextRector;
+use DrupalRector\Drupal9\Rector\Deprecation\AssertOptionSelectedRector;
+use DrupalRector\Drupal9\Rector\Deprecation\ConstructFieldXpathRector;
+use DrupalRector\Drupal9\Rector\Deprecation\GetAllOptionsRector;
+use DrupalRector\Drupal9\Rector\Deprecation\GetRawContentRector;
+use DrupalRector\Drupal9\Rector\Deprecation\PassRector;
+use DrupalRector\Drupal9\Rector\Deprecation\UiHelperTraitDrupalPostFormRector;
+use DrupalRector\Drupal9\Rector\Deprecation\UserPasswordRector;
+use DrupalRector\Drupal10\Rector\ValueObject\AssertLegacyTraitConfiguration;
```
bbrala: This looks wrong

mglaman (approved):

PR #269, opened 2023-11-22
Link packagist version shield to package on packagist.org

kasperg:
## Description

This seems to be a pretty standard approach. Other projects using the same approach include:

- https://github.com/laravel/laravel
- https://github.com/sebastianbergmann/phpunit

## To Test
- Check the rich diff in the PR
- See that the Packagist shield links to the package on Drupal.org.
   
## Drupal.org issue
This seems to be GitHub only. Is a Drupal.org issue necessary here?

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ Automate fixing deprecated Drupal code.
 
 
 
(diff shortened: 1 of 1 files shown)


bbrala:
Thanks for the help :)

bbrala (approved):

PR #273, opened 2023-11-25
PHPUnit: Rector to fix missing parent::setUp and parent::tearDown methods

bbrala:
## Description
Based on this phpstan rule: https://github.com/phpstan/phpstan-phpunit/commit/d686c82965e4e44fb753ca4342fcf797c1f01c88

Adds parent::setUp and parent::tearDown to those methods when missing.

## To Test
Run tests

### DIFF
--- a/src/Rector/PHPUnit/ShouldCallParentMethodsRector.php
+++ b/src/Rector/PHPUnit/ShouldCallParentMethodsRector.php
@@ -43,12 +43,6 @@ class ShouldCallParentMethodsRector extends AbstractScopeAwareRector
             return null;
         }
 
-        $parentClassesNames = $scope->getClassReflection()
-            ->getParentClassesNames();
-        if (!in_array(\PHPUnit\Framework\TestCase::class, $parentClassesNames)) {
-            return null;
-        }
-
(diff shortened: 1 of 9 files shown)


src/Rector/PHPUnit/ShouldCallParentMethodsRector.php
```diff
@@ -0,0 +1,166 @@
...
+
+        if ($parentClass === null) {
+            return null;
+        }
+
+        $parentClassesNames = $scope->getClassReflection()
+            ->getParentClassesNames();
+        if (!in_array(\PHPUnit\Framework\TestCase::class, $parentClassesNames)) {
+            return null;
+        }
```
mglaman: Doesn't the subclass check beforehand catch this? Or is it part of what was copy pasted

bbrala: I was scared for not finding deeper subclasses, but didn't test without this. Will run a quick test

bbrala: Seems you are actually right :x Also added a test that shows it doesn't remove it without a testcase as parent

agentrickard (approved):
Looks clean. Makes sense.

PR #280, opened 2023-12-28
Hotfix: New rector FILE_STATUS_PERMANENT was not added to deprecation list. …

bbrala:
…Only config object was intialized.

## Description

## To Test

## Drupal.org issue

### DIFF
--- a/src/Drupal8/Rector/Deprecation/DBRector.php
+++ b/src/Drupal8/Rector/Deprecation/DBRector.php
@@ -61,7 +61,7 @@ class DBRector extends AbstractRector implements ConfigurableRectorInterface
     private array $configuration;
 
     /**
-     * @var \DrupalRector\Services\AddCommentService
+     * @var AddCommentService
      */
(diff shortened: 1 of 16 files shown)


src/Drupal8/Rector/Deprecation/DBRector.php
```diff
@@ -61,7 +61,7 @@ class DBRector extends AbstractRector implements ConfigurableRectorInterface
     private array $configuration;
 
     /**
-     * @var \DrupalRector\Services\AddCommentService
+     * @var AddCommentService
```
agentrickard, line 64: My understanding is that @var descriptions should use the full class path, so I don't understand these changes.

agentrickard: This is part of the Symfony code standards. https://cs.symfony.com/doc/rules/import/fully_qualified_strict_types.html

agentrickard (approved):

PR #284, opened 2024-01-11
feat: upgrade rector, current rector is broken with latest phpstan release

bbrala:
## Description

See: https://github.com/rectorphp/rector/issues/8395

## To Test
Run tests <3 

## Drupal.org issue
no.

PR #285, opened 2024-01-11
feat: Conflict phpstan >1.10.54, currently rector is broken with latest phpstan

bbrala:
## Description
See: https://github.com/rectorphp/rector/issues/8395

Work on upgrade here: #284

## To Test
Run tests

## Drupal.org issue


bbrala:
Also did a fixstyle since that was needed.

Also needed to change the workflow a little to handle conflict properly now that phpstan is a core dependency.

bbrala:
We can actually fix this.

PR #288, opened 2024-01-16
Fix: double deprecated calls when already refactored

bbrala:
## Description
This will fix the double deprecated calls. In the future we might need to refactor this, since something might actually need a nested deprecated call. That would be kinda nasty, but is an edge case that could happen (tm).

Codestyle is a bit noisy, annoying, but ah well.

## To Test
Run test.

## Drupal.org issue
https://www.drupal.org/project/rector/issues/3414169

### DIFF
--- a/src/Rector/AbstractDrupalCoreRector.php
+++ b/src/Rector/AbstractDrupalCoreRector.php
@@ -8,7 +8,9 @@ use Drupal\Component\Utility\DeprecationHelper;
 use DrupalRector\Contract\VersionedConfigurationInterface;
 use PhpParser\Node;
 use PhpParser\Node\Expr\ArrowFunction;
+use PHPStan\Reflection\MethodReflection;
 use Rector\Contract\Rector\ConfigurableRectorInterface;
+use Rector\NodeTypeResolver\Node\AttributeKey;
 use Rector\Rector\AbstractRector;
 
 abstract class AbstractDrupalCoreRector extends AbstractRector implements ConfigurableRectorInterface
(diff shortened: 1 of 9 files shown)


src/Rector/AbstractDrupalCoreRector.php
```diff
@@ -29,6 +31,31 @@ public function configure(array $configuration): void
...
+            return false;
+        }
+        if ($function->getName() !== 'backwardsCompatibleCall'
+            || $function->getDeclaringClass()->getName() !== DeprecationHelper::class
+        ) {
+            return false;
+        }
+
+        return $parameter !== null && $parameter->getName() === 'deprecatedCallable';
+    }
```
mglaman, line 57: Does Rector have a node visitor to just stop processing so that it doesn't handle nodes in this execution tree?

bbrala: Think so yes. My old fix for phpstan would probably also work. This seemed easier is 0retty much why I used this.

mglaman: I guess we're calling it in our base class, which all rules should extend. This is fine, but could lead to a "bug" if a rule didn't. But probably a minimal problem

mglaman (approved):

src/Rector/AbstractDrupalCoreRector.php
bbrala: valid point. But if it doesn't extend this class, it should not add the deprecation helper. But you are right though, it could slip through if we need some sort of exotic way to do things. Still would probably use the base class for the helpers.

PR #289, opened 2024-01-24
feat: add rector rule for deprecated GDToolkit 

timohuisman:
feat: add rector rule for deprecated GDToolkit resource methods and properties in 10.2

## Description
GDToolkit::setResource(), GDToolkit::getResource() and GDToolkit::resource are deprecated.

GDToolkit has been updated to use the GD extension classed object \GdImage instead of a generic resource.

## Todo
- There should probably be a test for `RenamePropertyRector`, but I'm not sure where to start.

## To Test
- composer test

## Drupal.org issue
https://www.drupal.org/project/rector/issues/3416757
https://www.drupal.org/node/3265963

### DIFF
--- a/config/drupal-10/drupal-10.2-deprecations.php
+++ b/config/drupal-10/drupal-10.2-deprecations.php
@@ -7,8 +7,6 @@ use DrupalRector\Rector\Deprecation\MethodToMethodWithCheckRector;
 use DrupalRector\Rector\ValueObject\FunctionToStaticConfiguration;
 use DrupalRector\Rector\ValueObject\MethodToMethodWithCheckConfiguration;
 use Rector\Config\RectorConfig;
-use Rector\Renaming\Rector\PropertyFetch\RenamePropertyRector;
-use Rector\Renaming\ValueObject\RenameProperty;
 
 return static function (RectorConfig $rectorConfig): void {
     // https://www.drupal.org/node/2999981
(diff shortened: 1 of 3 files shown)


timohuisman:
The tests are failing because of https://github.com/palantirnet/drupal-rector/pull/290

config/drupal-10/drupal-10.2-deprecations.php
```diff
@@ -3,12 +3,27 @@
...
     ]);
+
+    // https://www.drupal.org/node/3265963
+    $rectorConfig->ruleWithConfiguration(MethodToMethodWithCheckRector::class, [
+        new MethodToMethodWithCheckConfiguration('Drupal\system\Plugin\ImageToolkit\GDToolkit', 'getResource', 'getImage'),
+        new MethodToMethodWithCheckConfiguration('Drupal\system\Plugin\ImageToolkit\GDToolkit', 'setResource', 'setImage'),
+    ]);
+
+    // https://www.drupal.org/node/3265963
+    $rectorConfig->ruleWithConfiguration(RenamePropertyRector::class, [
```
bbrala: Unfortunately this does not support backwards compatible fixes. We either need to pull that code into a class that extends our base class, or perhaps think up a wrapper that can wrap external Rector classes. Not sure how feasable that is though. 

I can imagine something like a class that allows for a reference to the rector as a configuration object, this could then be used to perhaps get a node types. This might break though if you try and wrap multiple (in config) since those may have different node types they apply to, resulting in breakage.

bbrala (changes requested):

bbrala:
Wonder if renaming ::resource is even something we should do. I mean, its a protected var, you might use it in an extending class but you could argue that is not common at all.

When looking at contrib i see no instances of that. 

https://git.drupalcode.org/search?group_id=2&page=4&repository_ref=11.x&scope=blobs&search=extends+GDToolkit

So perhaps just remove the property fetch and only do the public API.

bbrala (approved):

PR #296, opened 2024-03-09
feat: Optimize AnnotationToAttributeRector

bbrala:
## Description
Add some better handling of types using Doctrine annotation service. Also add test for ContentEntityType and ConfigEntityType annotation. This seems like it will handle most things in core (with support for BC fixes for contrib).

Also added documentation for core plugin conversion so it can be started to be used.

Lets get this merged :)

## To Test
- Run tests
- Try a plugin conversion? ; ) 

## Drupal.org issue
https://www.drupal.org/project/rector/issues/3265946


bbrala:
Also this was used here to generate a MR: https://www.drupal.org/project/drupal/issues/3396166

docs/core_plugin_conversion.md
```diff
@@ -0,0 +1,88 @@
...
+    new \DrupalRector\Drupal10\Rector\ValueObject\AnnotationToAttributeConfiguration('10.0.0', '10.0.0', 'ContentEntityType', 'Drupal\Core\Entity\Attribute\ContentEntityType'),
+    new \DrupalRector\Drupal10\Rector\ValueObject\AnnotationToAttributeConfiguration('10.0.0', '10.0.0', 'ConfigEntityType', 'Drupal\Core\Entity\Attribute\ConfigEntityType'),
+  ]);
+
+  $rectorConfig->autoloadPaths([
+    './lib',
+    './modules',
+    './profiles',
+    './themes'
+  ]);
```
mglaman: These settings and you're in `core` but the script execution looks like it's from the directory above

mglaman (approved):

docs/core_plugin_conversion.md
bbrala: You're right

PR #304, opened 2024-05-28
New rector (10.3): Add rule for file_icon_class() and file_icon_map()

timohuisman:
## Description
`file_icon_class()` and `file_icon_map()` are deprecated and replaced with static methods on `\Drupal\file\IconMimeTypes`. I've used the existing `FunctionToStaticConfiguration` to replace the deprecated functions.

@bbrala mentioned that `stubs/Drupal/Drupal.php` should be upped to `10.99.x-dev` because we always need to run the tests in this major.

## To Test
- run `composer test`

## Drupal.org issue
https://www.drupal.org/project/rector/issues/3450324
https://www.drupal.org/node/3411269


bbrala:
For now its ok to add those errors in phpstan to the baseline, we'll need to fix those, so perhaps a followup on the main issue on d.o. is good :)

bbrala (approved):
failure is completely unrelated. Thanks!

PR #312, opened 2024-12-12
feat: Add support for PHPstan 2 and Rector 2

ptmkenny:
## Description
Updates to Rector 2 + PHPStan 2

## To Test
- Not ready yet; Rector 2 has not been released (currently rc3; wanted to get this going to test how much work is needed to update)

## Drupal.org issue
https://www.drupal.org/project/rector/issues/3493467


holo96:
Rector has stable release, can we merge this now?

bbrala:
This could be merged, but needs updating i think in regards to the version numbers it seems, we'd want to work from stable right?

holo96:
Yes, and I think PHP ^7.4 is supported, so change in .github/workflows/functional_test__rector_examples.yml is not needed
Also releasing Drupal rector 2.0 should be considered, following [rector](https://github.com/rectorphp/rector) 2.0 and also [phpstan-drupal](https://github.com/mglaman/phpstan-drupal) 2.0

ptmkenny:
I bumped the rector dependency to `^2.0`.

I agree with @Holo96 that it would be nice to have drupal-rector 2.0 to match phpstan-drupal 2.0 and rector 2.0.

Maintainers, please feel free to edit this MR however you like; as stating in the first post, I opened this as an experiment to test the feasibility of moving to phpstan 2.0.

bbrala:
Ok, i did some things to make compativbility happen.

Right now it tests on 1 and 2, since it seems that might juist be possibel. Unfortunately it seems that AnnotationToAttribute rector breaks in phpstan 2 right now. There might be more breaking tests.

I'm considering also to drop Drupal8 support, but not sure yet if that is needed.

bbrala:
Fixed a few more things thta were weird. Only real issue remaining is the broken AnnotationToAttributeRector

config/drupal-9/drupal-9.0-deprecations.php
```diff
@@ -15,14 +15,20 @@
         return new AddCommentService();
     });
 
+    if (defined('TwigSetList::TWIG_24')) {
+        $twigSet = TwigSetList::TWIG_24;
+    } else {
+        $twigSet = TwigSetList::TWIG_240;
+    }
```
holo96: @bbrala 
This does not work for me, I needed whole namespace, there is same issue in `config/drupal-10/drupal-10.0-deprecations.php`
```suggestion
    if (defined('\Rector\Symfony\Set\TwigSetList::TWIG_24')) {
        $twigSet = TwigSetList::TWIG_24;
    } else {
        $twigSet = TwigSetList::TWIG_240;
    }
```

config/drupal-9/drupal-9.0-deprecations.php
```diff
@@ -15,14 +15,20 @@
         return new AddCommentService();
     });
 
+    if (defined('\Rector\Symfony\Set\TwigSetList::TWIG_24')) {
+        $twigSet = TwigSetList::TWIG_24;
+    } else {
+        $twigSet = TwigSetList::TWIG_240;
+    }
```
holo96: Also `use Rector\Exception\ShouldNotHappenException;` should be added.

```suggestion
    if (defined(TwigSetList::class .'::TWIG_24')) {
        $twigSet = TwigSetList::TWIG_24;
    } elseif (defined(TwigSetList::class .'::TWIG_240'))  {
        $twigSet = TwigSetList::TWIG_240;
    } else {
        throw new ShouldNotHappenException('Could not detect twig set.');
    }
```

config/drupal-10/drupal-10.0-deprecations.php
```diff
@@ -9,6 +9,12 @@
 use Rector\Symfony\Set\TwigSetList;
 
 return static function (RectorConfig $rectorConfig): void {
+    if (defined('TwigSetList::TWIG_24')) {
+        $twigSet = TwigSetList::TWIG_24;
+    } else {
+        $twigSet = TwigSetList::TWIG_240;
+    }
```
holo96: Also `use Rector\Exception\ShouldNotHappenException;` should be added.
```suggestion
if (defined(TwigSetList::class .'::TWIG_24')) {
    $twigSet = TwigSetList::TWIG_24;
  } elseif (defined(TwigSetList::class .'::TWIG_240'))  {
    $twigSet = TwigSetList::TWIG_240;
  } else {
    throw new ShouldNotHappenException('Could not detect twig set.');
  }
```

src/Drupal8/Rector/Deprecation/DrupalSetMessageRector.php
```diff
@@ -84,7 +84,6 @@ public function refactor(Node $node): ?Node
         }
 
         $expression = $node->expr;
-        assert($expression instanceof Node\Expr\FuncCall);
```
agentrickard, line 87: **question:** Why was this check removed?

src/Drupal9/Rector/Deprecation/PassRector.php
```diff
@@ -51,7 +57,12 @@ public function refactor(Node $node)
         }
 
         if ($this->getDeclaringSource($node->expr) === 'Drupal\KernelTests\AssertLegacyTrait') {
-            return NodeTraverser::REMOVE_NODE;
+            if (defined('\PhpParser\NodeVisitor::REMOVE_NODE')) {
+                return NodeVisitor::REMOVE_NODE;
+            } else {
+                /* @phpstan-ignore-next-line */
```
agentrickard, line 63: **question:** Why and ignore here and not on line 61? Should we instead make this an `elseif` on line 62 and type check?

src/Rector/Deprecation/MethodToMethodWithCheckRector.php
```diff
@@ -117,8 +117,6 @@ public function refactor(Node $node): ?Node
 
     public function refactorNode(Node\Expr\MethodCall $node, ?Node\Stmt\Expression $statement, MethodToMethodWithCheckConfiguration $configuration): ?Node\Expr\MethodCall
     {
-        assert($node instanceof Node\Expr\MethodCall);
```
agentrickard, line 120: **note:** Here I can see why the assert was removed, since its redundant to the method definition.

agentrickard (commented):
This is looking very good. 

I did get some additional phpstan errors locally, but I am running php 8.4. We likely need a new ticket / plan for that update.

src/Drupal8/Rector/Deprecation/DrupalSetMessageRector.php
bbrala: Because a few lines above:

```
        if (!($node->expr instanceof Node\Expr\FuncCall)) {
            return null;
        }
```

src/Drupal9/Rector/Deprecation/PassRector.php
bbrala: we might be able to, but this is jsut BC, this constant does exist in 2.0 though, but is deprecated, so it gets kinda confused. 

If i do an elseif and then an exception, i do get an error also in phpstan in version 2.

bbrala:
Hmmz, while fixing up i noticed SymfonySetList has just been deprecated, fun cause that cant be fixed in 1.0.

I'd say we should release this, and perhaps then split 1.0 and 2.0, with 2.0 only supporting rector 2.0 and start removing old rectors for Drupal 8 and 9. That would make life a lot easier i think and those are unsupported for a while now.

We could try and keep up in 1.0 if we add a rector that is for Drupal 10, but i dont thing we have many big ones left from 10.0. Its about time we start running on 11.x and start making some rectors there.

holo96:
I did some manual testing, everything works great!

I would appreciate release, because this is blocking me from updating rector & phpstan (which blocks some php 8.4 features), and there are same useful features in here, like preventing someone to write annotations

bbrala (approved):

PR #313, opened 2025-01-24
fix: remove deprecated SetInterface

bbrala:
## Description
SetInterface has been deprecated, this needs removing. We could move to the new SetProvider to supply sets eventually. Which also enables things like composer based sets. For now, just fix the deprecation. This should help us make the transition to 2.0 easier.

## To Test
- Tests should be grean

## Drupal.org issue
No issue.


bbrala:
Closing this, will do a single deprecations pr

PR #316, opened 2025-05-14
chore: Remove use of protected property NodeNameResolver on HookConvertRector

samsonasik:
## Description

Rector is removing use of protected property `nodeNameResolver` in rules, use `isName()` directly instead. Ref PR:

- https://github.com/rectorphp/rector-src/pull/6875

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -137,6 +137,16 @@ parameters:
 
 		-
 			message: '''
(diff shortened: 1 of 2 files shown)


samsonasik:
Ready to merge 👍 

phpstan notice is unrelated.

bbrala:
Did fix the little baseline error, seems phpstan improved. 

Thank you for your help! :)

bbrala (approved):

PR #317, opened 2025-05-24
Fix method not found on GetDeclaringSourceTrait::getDeclaringSource()

samsonasik:
## Description

We got various reported issues because of this, see

https://github.com/rectorphp/rector/issues?q=is%3Aissue%20state%3Aclosed%20drupal-rector

this ensure only get method when exists to avoid error:

```
     "System error: "Method assertEscaped() was not found in reflection of class                                    
     Drupal\Tests\views_aggregator\Functional\Plugin\ViewsAggregatorStyleTableTest."  
```

For note: I think you better open "Issues" tab so people can report here as well


samsonasik:
@bbrala ready to merge 👍

bbrala (approved):

PR #318, opened 2025-06-22
Update Deny list to include to missing hooks and remove preprocess

nlighteneddesign:
## Description
There are two procedural only hooks we missed.

Also preprocess is now supported for 11.2.

## To Test
Needs manual testing.

## Drupal.org issue
https://www.drupal.org/project/drupal/issues/3524872

### DIFF
--- a/src/Rector/Convert/HookConvertRector.php
+++ b/src/Rector/Convert/HookConvertRector.php
@@ -192,7 +192,9 @@ CODE_SAMPLE
             $infoFile = reset($info);
             $this->module = basename($infoFile, '.info.yml');
             $filename = pathinfo($this->file->getFilePath(), \PATHINFO_FILENAME);
-            $hookClassName = ucfirst(CaseStringHelper::camelCase(str_replace('.', '_', $filename).'_hooks'));
+            // Remove _theme before merge.
+            // This is important.
+            $hookClassName = ucfirst(CaseStringHelper::camelCase(str_replace('.', '_', $filename).'_theme_hooks'));
             $counter = '';
             do {
                 $candidate = "$hookClassName$counter";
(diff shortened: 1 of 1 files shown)


src/Rector/Convert/HookConvertRector.php
ms-slasher13, line 1: ` || preg_match('/^(post_update_|update_\d+$)/', $hook)` is missing in line 258. see `checkForProceduralOnlyHooks` in https://git.drupalcode.org/project/drupal/-/blob/11.2.x/core/lib/Drupal/Core/Hook/HookCollectorPass.php?ref_type=heads#L585

src/Rector/Convert/HookConvertRector.php
```diff
@@ -241,17 +243,19 @@ protected function createMethodFromFunction(Function_ $node): ?ClassMethod
...
+              'module_implements_alter',
+              'removed_post_updates',
+              'requirements',
+              'schema',
+              'uninstall',
+              'update_dependencies',
+              'update_last_removed',
             ];
-            if (in_array($hook, $procOnly) || str_starts_with($hook, 'preprocess') || str_starts_with($hook, 'process')) {
+            if (in_array($hook, $procOnly)) {
```
ms-slasher13, line 256: ` || preg_match('/^(post_update_|update_\d+$)/', $hook)` is missing in line 258. see `checkForProceduralOnlyHooks` in https://git.drupalcode.org/project/drupal/-/blob/11.2.x/core/lib/Drupal/Core/Hook/HookCollectorPass.php?ref_type=heads#L585

nlighteneddesign: It is not missing, this script does not run on .install or post_update files so those hooks will not be affected.

ms-slasher13: These hooks aren't always in .install or post_update files.

nlighteneddesign: Can you share an example?

bbrala:
All good, i'll fix php cs issues in the next issue.

bbrala (approved):

PR #320, opened 2025-10-20
Better replacement for dynamic uppercase parts in Implements hook_xy

Berdir:
## Description

Replacement of hooks that contain a dynamic part such as FORM_ID or ENTITY_TYPE currently don't work.

The problem is that it explicitly only replaces one section of uppercase characters with a regex, so it will try to match against an explicit ID or TYPE. Adding multiple dynamic sections wouldn't work either because the actual replacement might be just one part.

I suggest replacing the whole thing with a preg_replace() that finds any complete section of uppercase characters.

## To Test
- Run it on a module that has such a hook.

## Drupal.org issue
There is no issue since this is a bugfix to an existing rule.

### DIFF
--- a/src/Rector/Convert/HookConvertRector.php
+++ b/src/Rector/Convert/HookConvertRector.php
@@ -320,25 +320,11 @@ CODE_SAMPLE
     {
         // If the doxygen contains "Implements hook_foo()" then parse the hook
         // name. A difficulty here is "Implements hook_form_FORM_ID_alter".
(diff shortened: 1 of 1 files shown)


nlighteneddesign:
I tested this and this works!

nlighteneddesign (approved):
I tested locally with the metatag module and it works as expected

bbrala (approved):

PR #326, opened 2026-04-08
Allow rector ^2.4.1 and replace file with getFile() if method exists

samsonasik:
## Description

In rector 2.4.1, the `file` property usage is deprecated, use `->getFile()` instead.

- https://github.com/rectorphp/rector-src/pull/7962

since this repo allow rector 1, this need method_exists() check.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,7 @@
         "ast"
     ],
     "require": {
-        "rector/rector": "^1 || ^2",
+        "rector/rector": "^1 || ^2.4.1",
         "webflo/drupal-finder": "^1.2"
     },
     "license": "MIT",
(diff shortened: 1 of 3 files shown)


samsonasik:
@bbrala ready to merge 👍

samsonasik:
CI notice is unrelated :),  ready to merge 👍

composer.json
```diff
@@ -9,7 +9,7 @@
         "ast"
     ],
     "require": {
-        "rector/rector": "^1 || ^2",
+        "rector/rector": "^1 || ^2.4.1",
```
bbrala: Why this extra specification?

samsonasik: The `getFile()` method exists since 2.4.1, feel free to revert and take care the rest, this just ensure it can see when the requirement needed

PR #341, opened 2026-06-01
Fix D9+D11 set collision and document running against Drupal 10

bbrala:
## Why

Teams on **Drupal 10** want to run Drupal Rector to clean up deprecations *before* upgrading to Drupal 11. Investigating how to support that surfaced a real crash in the bundled configuration, plus a packaging gotcha for sites that pin PHPStan 1.

### `fix:` D9+D11 set collision (container-build crash)

Loading the Drupal 9 **and** Drupal 11 sets together crashed before processing any file:

```
InvalidArgumentException: Each configuration item must be an instance of
"DrupalRector\Drupal9\Rector\ValueObject\FunctionToFirstArgMethodConfiguration"
```

`Drupal9\…\FunctionToFirstArgMethodRector` (and `Drupal8\…\DrupalServiceRenameRector`) subclass their generic rules. Rector applies configuration via illuminate-container `afterResolving` callbacks that match by **`instanceof`**, so the subclass instances also received the generic rule's version-tagged configuration and their strict `configure()` type guard threw.

The bundled `rector.php` loads all four sets, so the **default config crashed on any run**. It matters for D10 specifically because some D10-era deprecations live only in the D11 set (e.g. `LoadAllIncludesRector`, deprecated in 10.3).

**Fix:** the subclasses filter `configure()` to their own configuration type and ignore the rest; the generic rule instance still applies its own configs, so no coverage is lost. Added a regression test that registers both rules together and fails on the pre-fix code.

### `docs:` Running against a Drupal 10 project

New guide (`docs/running-against-drupal-10.md`, linked from the README) covering:
- **Direct install** for plain D10 sites.
- A **standalone-runner** recipe for sites whose PHPStan 1 tooling (e.g. `mglaman/phpstan-drupal`) conflicts with Rector 2's PHPStan 2 requirement — a separate Composer project loading the per-version sets and pointing at the site via `--autoload-file`.

## Validation

- Full suite green (**510 tests**, +1 regression), PHPStan clean.
- Verified end-to-end against a real Drupal 10.6 site: D9, D10.2 and D10.3 deprecations all transform with the full D8–D11 set loaded, both for a direct install and via the standalone runner.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,12 @@ release-by-release.
 
 ## [Unreleased]
 
+### Added
+
+- Guide: [Running against a Drupal 10 project](docs/running-against-drupal-10.md) — covers the
+  direct install and a standalone-runner recipe for sites whose PHPStan 1 tooling conflicts with
+  Rector 2's PHPStan 2 requirement.
+
 ### Fixed
 
 - Loading the Drupal 9 and Drupal 11 sets together no longer crashes at
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,17 @@ release-by-release.
 
 ## [Unreleased]
 
+### Fixed
+
+- Loading the Drupal 9 and Drupal 11 sets together no longer crashes at
+  container-build time. The Drupal 9 `FunctionToFirstArgMethodRector` (and the
+  Drupal 8 `DrupalServiceRenameRector`) subclass the generic rule, so Rector
+  delivered the generic rule's configuration to the subclass instance as well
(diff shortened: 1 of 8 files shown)


miiimooo:
I followed this guide, but I it runs into problems. 

>  [ERROR] Could not detect twig set.                                                                                     

I can trace the error to ~/drupal-rector-runner/vendor/palantirnet/drupal-rector/config/drupal-9/drupal-9.0-deprecations.php and ~/drupal-rector-runner/vendor/palantirnet/drupal-rector/config/drupal-10/drupal-10.0-deprecations.php

My site in a separate directory is on core 10.6.15

PR #358, opened 2026-06-04
feat(11.3): cover remaining template_preprocess_*() deprecations (#3504125)

bbrala:
## What

Completes `FunctionToServiceRector` coverage of the `template_preprocess_*()` family deprecated in **drupal:11.3.0** (removed in 12.0.0), change record [#3504125](https://www.drupal.org/node/3504125).

Previously only the 11.2 members (`time`, `datetime_form`, `datetime_wrapper`, `links`, `container`, `html`, `page`) plus `layout` (11.3) were configured. This adds the **14 remaining** 11.3 functions, each mapped to its `*Preprocess` service method:

| Function(s) | Service |
|---|---|
| `table`, `tablesort_indicator`, `item_list`, `region`, `maintenance_page`, `maintenance_task_list`, `install_page` | `Drupal\Core\Theme\ThemePreprocess` |
| `image` | `Drupal\Core\Theme\ImagePreprocess` |
| `breadcrumb` | `Drupal\Core\Breadcrumb\BreadcrumbPreprocess` |
| `pager` | `Drupal\Core\Pager\PagerPreprocess` |
| `field`, `field_multiple_value_form` | `Drupal\Core\Field\FieldPreprocess` |
| `menu_local_task`, `menu_local_action` | `Drupal\Core\Menu\MenuPreprocess` |

## Why

Running against contrib (e.g. `jsonapi_extras`, which calls `template_preprocess_table()` from a custom preprocess function) surfaced this gap — the function is deprecated but had no rector coverage.

## Notes

- No new rector code — these are config entries on the existing generic `FunctionToServiceRector`. Mappings, deprecation version, and CR were read directly from core's deprecated function bodies (`core/includes/theme.inc`).
- Calls are **BC-wrapped via `DeprecationHelper::backwardsCompatibleCall()`** and version-gated to `11.3.0`, identical to the existing `template_preprocess_*` entries, so generated code is safe on older minors.
- `template_preprocess_authorize_report()` is **intentionally excluded** — its deprecation states "There is no replacement" (manage the code via composer), so it is not a function-to-service candidate.

## Testing

- New fixture `template_preprocess_3504125.php.inc` exercises all 14 functions (before → BC-wrapped after).
- `FunctionToServiceRectorTest`: **31/31 pass**.
- `composer check-style` and `composer phpstan`: clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,17 @@ release-by-release.
 
 ### Added
 
+- **`template_preprocess_*()` family (Drupal 11.3, [#3504125](https://www.drupal.org/node/3504125))** —
+  extended `FunctionToServiceRector` coverage with 14 more `template_preprocess_*()`
+  functions deprecated in drupal:11.3.0 and removed in drupal:12.0.0. Each is
+  rewritten (BC-wrapped via `DeprecationHelper`) to the equivalent `*Preprocess`
+  service method: `table`, `tablesort_indicator`, `item_list`, `region`,
+  `maintenance_page`, `maintenance_task_list`, `install_page` → `ThemePreprocess`;
+  `image` → `ImagePreprocess`; `breadcrumb` → `BreadcrumbPreprocess`;
+  `pager` → `PagerPreprocess`; `field`, `field_multiple_value_form` →
+  `FieldPreprocess`; `menu_local_task`, `menu_local_action` → `MenuPreprocess`.
+  (`template_preprocess_authorize_report()` has no replacement and is
+  intentionally excluded.)
 - **`RemoveDrupalToStringTraitRector`** — removes
   `use Drupal\Component\Utility\ToStringTrait;` from a class body and inserts
   an inline `public function __toString(): string { return (string)
--- a/config/drupal-11/drupal-11.3-deprecations.php
+++ b/config/drupal-11/drupal-11.3-deprecations.php
@@ -69,10 +69,26 @@ return static function (RectorConfig $rectorConfig): void {
     ]);
 
     // https://www.drupal.org/node/3504125
-    // template_preprocess_layout() deprecated in drupal:11.3.0, removed in drupal:12.0.0.
-    // Replaced by \Drupal\layout_discovery\Hook\LayoutDiscoveryThemeHooks::preprocessLayout().
+    // The template_preprocess_*() functions deprecated in drupal:11.3.0,
+    // removed in drupal:12.0.0. Initial template_preprocess functions are now
+    // registered directly in hook_theme(); each is replaced by the equivalent
+    // *Preprocess service method.
     $rectorConfig->ruleWithConfiguration(FunctionToServiceRector::class, [
         new FunctionToServiceConfiguration('11.3.0', 'template_preprocess_layout', 'Drupal\layout_discovery\Hook\LayoutDiscoveryThemeHooks', 'preprocessLayout', true),
+        new FunctionToServiceConfiguration('11.3.0', 'template_preprocess_table', 'Drupal\Core\Theme\ThemePreprocess', 'preprocessTable'),
+        new FunctionToServiceConfiguration('11.3.0', 'template_preprocess_tablesort_indicator', 'Drupal\Core\Theme\ThemePreprocess', 'preprocessTablesortIndicator'),
+        new FunctionToServiceConfiguration('11.3.0', 'template_preprocess_item_list', 'Drupal\Core\Theme\ThemePreprocess', 'preprocessItemList'),
+        new FunctionToServiceConfiguration('11.3.0', 'template_preprocess_region', 'Drupal\Core\Theme\ThemePreprocess', 'preprocessRegion'),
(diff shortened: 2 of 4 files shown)


bbrala:
## Code Review

### Overview
Adds 14 config entries to the existing generic `FunctionToServiceRector`, completing coverage of the `template_preprocess_*()` functions deprecated in drupal:11.3.0 / removed in 12.0.0. No new rector code — pure configuration plus a fixture.

### Correctness — verified against core
Checked every mapping against `drupal-core` 11.x (`core/includes/theme.inc`). All 14 are **exact matches** to what core's own deprecated function bodies forward to:

- ✅ Service IDs (`ThemePreprocess`, `ImagePreprocess`, `BreadcrumbPreprocess`, `PagerPreprocess`, `FieldPreprocess`, `MenuPreprocess`) — core itself uses `\Drupal::service(ThemePreprocess::class)`, so the IDs are proven valid.
- ✅ All 14 method names match.
- ✅ Version `11.3.0` and CR `#3504125` confirmed in the `@deprecated` / `@trigger_error` annotations.
- ✅ Exclusion of `template_preprocess_authorize_report()` (and `_field_multiple_value_form_sort_helper()`) is correct — both state "no replacement."

### Conventions & BC
- ✅ Matches the established `template_preprocess_layout` entry style.
- ✅ BC-wrapped via `DeprecationHelper::backwardsCompatibleCall()`, version-gated to `11.3.0` — safe on older minors. No silent-divergence risk: the fallback simply calls the still-present deprecated function.
- ✅ CHANGELOG entry present and accurate.

### One thing worth noting (not a blocker)
`template_preprocess_field(&$variables, $hook)` takes **two** parameters in core, but `FieldPreprocess::preprocessField(array &$variables)` takes **one** — core's deprecated body itself drops `$hook`. The rector forwards `$node->args` verbatim, so a real call site `template_preprocess_field($variables, $hook)` rewrites to `->preprocessField($variables, $hook)` with an extra arg. PHP silently ignores surplus positional args, so this is **not** a fatal — but the generated code carries a vestigial `$hook` the method ignores. The fixture only exercises the single-arg form. Optional: add a 2-arg fixture to document the shape.

### Tests
- ✅ `FunctionToServiceRectorTest`: **31/31 pass** (re-ran locally).
- Minor coverage gap: no fixture for the realistic 2-arg `template_preprocess_field($variables, $hook)` call shape.

### Security / Performance
No concerns — config-only change to an existing, well-tested rector.

### Verdict
**Approve.** Mappings are exact against core, BC handling is correct, tests green. The only observation (vestigial `$hook` arg on `template_preprocess_field`) is harmless in PHP and matches core's own forwarding intent.

mglaman (approved):
Config-only, correctly placed. All 14 `template_preprocess_*()` functions route through `FunctionToServiceRector`, which BC-wraps each call in `DeprecationHelper::backwardsCompatibleCall()` at 11.3.0, so the default `drupal-11.3-deprecations.php` set is right. The function→service-method mappings line up with the deprecated function bodies in core.

Excluding `template_preprocess_authorize_report()` because core ships no replacement is the correct call, and documenting it beats a silent gap. Single fixture exercises all 14.

Approving.

bbrala:
## Review: ✅ Approve

Config-only change to the existing generic `FunctionToServiceRector`. I independently verified all 14 mappings against drupal-core (`11.x`).

### Correctness — verified against core
Checked each deprecated function body in `core/includes/theme.inc` against the PR's mappings — every one matches exactly:

- **Service ID = FQCN.** Core's BC bodies call `\Drupal::service(ThemePreprocess::class)->…`, so the rector's `\Drupal::service('Drupal\Core\Theme\ThemePreprocess')` output is correct.
- **Method names** all match the core delegations (`preprocessTable`, `preprocessTablesortIndicator`, …, `preprocessMenuLocalAction`).
- **Service classes** all match: `ThemePreprocess` (7), `ImagePreprocess`, `BreadcrumbPreprocess`, `PagerPreprocess`, `FieldPreprocess` (2), `MenuPreprocess` (2).
- **Version `11.3.0`** and **CR #3504125** match the `@trigger_error` strings.
- **`authorize_report` exclusion is correct** — different CR (#3522119, deprecated 11.2.0), "There is no replacement," rightly out of scope.

### Conventions
- Matches the existing `template_preprocess_layout` entry style; the `true` configurable-target flag is correctly omitted since these targets are concrete core services (mirroring core's own delegation).
- Config and test `configured_rule.php` kept in sync.
- CHANGELOG entry present and well-formatted.

### Minor note (non-blocking)
`template_preprocess_field(&$variables, $hook)` takes a second `$hook` arg, but core's BC body drops it (`preprocessField($variables)`) and the service method signature is `preprocessField(array &$variables)`. `FunctionToServiceRector` copies `$node->args` verbatim (`FunctionToServiceRector.php:68`), so a `template_preprocess_field($variables, $hook)` call would generate `->preprocessField($variables, $hook)`. PHP silently ignores extra args on non-variadic methods, so this is **not fatal and behaviorally identical** — just a slight divergence from core's hand-written shim. Optional: add a two-arg `template_preprocess_field` line to the fixture to lock in that behavior.

### Tests
Fixture `template_preprocess_3504125.php.inc` exercises all 14 functions through the full BC-wrapped transformation; 31/31 passing, clean style/phpstan. Consistent with a config-only change.

**No blockers — safe to merge.**

PR #361, opened 2026-06-10
fix(Drupal11): handle isset()/unset() in ReplaceEntityOriginalPropertyRector

bbrala:
## Problem

`ReplaceEntityOriginalPropertyRector` rewrites `$entity->original` to the `$entity->getOriginal()` method call. But inside `isset()` / `unset()`, that produced **invalid PHP** — e.g. `isset($entity->getOriginal())` — because those constructs accept only a variable, not the result of an expression. The previous version simply skipped them, which also left deprecated code that silently breaks on Drupal 12.

## Fix

Mirror `EntityBase`'s magic methods ([CR 3295826](https://www.drupal.org/node/3295826)):

| Source | Rewrite |
|---|---|
| `isset($entity->original)` | `$entity->getOriginal() !== NULL` (`__isset` returns `getOriginal()`) |
| `unset($entity->original)` | `$entity->setOriginal(NULL)` (`__unset` calls `setOriginal(NULL)`) |

Both are wrapped in `DeprecationHelper`. `Isset_` is an `Expr`, so the parent's automatic BC-wrapping applies. `Unset_` is a `Stmt`, so it's wrapped manually with `$entity->original = NULL` as the pre-11.2 path (`unset()` can't be an arrow-function body). In a multi-operand `unset()`, only the `->original` operand is rewritten; the rest stay in a residual `unset()`, and leading comments are preserved.

### Left untouched (no valid equivalent — would be a fatal)
- Nested chains: `isset($entity->original->field)`, `unset($entity->original->field)`
- Multi-operand `isset()`

### Still rewritten
- Array-key position: `isset($map[$entity->original])` (a normal expression slot)
- `empty($entity->original)` (`empty()` accepts expressions)

## Implementation notes

- Direct `isset`/`unset` operands are detected via Rector's `IS_ISSET_VAR` / `IS_UNSET_VAR` attributes (set by `ContextNodeVisitor`), matching `ArrayDimFetchToMethodCallRector`.
- Those attributes are only set on the *top-level* operand, so nested `->original` fetches are guarded with a chain tag instead.

## Tests
Added fixtures: `isset`, `unset`, `unset_multiple`, `empty`, `nested_isset_unset_no_change`, `isset_array_key_rewritten`, plus a below-version gate. Full suite green (617 tests), PHPStan and code style clean.

> Note: this branch currently also shows 2 unrelated `chore:` commits (`rector.php` packaging) that are not yet in `main`; they'll drop from the diff once they land there.


bbrala:
Fixed a messy bug surfaced while running.

PR #363, opened 2026-06-12
feat: add ReplaceUserOneTimeAuthFunctionsRector for #3581056

bbrala:
## Summary

Adds `ReplaceUserOneTimeAuthFunctionsRector` covering Drupal core issue [#3581056](https://www.drupal.org/node/3581056) (change record [#3581062](https://www.drupal.org/node/3581062)).

Three `user.module` functions were deprecated in **drupal:11.4.0** and are removed in **drupal:13.0.0**, replaced by the new `\Drupal\user\OneTimeAuthentication` service:

| Deprecated function | Replacement |
|---|---|
| `user_pass_rehash($account, $timestamp)` | `\Drupal::service(\Drupal\user\OneTimeAuthentication::class)->generateHmac($account, $timestamp)` |
| `user_pass_reset_url($account, $options)` | `…->generateOneTimeLoginUrl($account, $options)->toString()` |
| `user_cancel_url($account, $options)` | `…->generateCancelConfirmUrl($account, $options)->toString()` |

The two URL methods return a `\Drupal\Core\Url` object (the old functions returned strings), so the rewrite chains `->toString()`.

### Why BC-wrapped

The rector extends `AbstractDrupalCoreRector` and wraps each rewrite in `DeprecationHelper::backwardsCompatibleCall(\Drupal::VERSION, '11.4.0', …)`. The `OneTimeAuthentication` service does not exist on Drupal < 11.4, so an unconditional rewrite would fatal on older cores.

### Not handled

`user_mail_tokens()` (deprecated in the same issue) is intentionally excluded — it is used as a string callback (`'callback' => 'user_mail_tokens'`) rather than called directly, and its replacement needs a `BubbleableMetadata` argument that cannot be synthesised.

## Verification against core

The deprecation landed in core commit `a38f1d17` (committed 2026-06-08). I fetched that commit directly and confirmed the method names, signatures (`generateHmac`/`generateOneTimeLoginUrl`/`generateCancelConfirmUrl`), the 11.4.0→13.0.0 window, and that the old functions delegate to the service.

## Live test against real contrib

Ran the rector against 5 D11-compatible contrib modules — all transformed correctly, no false positives:

| Module | Function(s) exercised |
|---|---|
| tfa | `user_pass_rehash` (nested in `hash_equals`) |
| user_registrationpassword | `user_pass_rehash` (in `.module`) |
| change_pwd_page | `user_pass_rehash` (nested in `hash_equals`) |
| rest_password | `user_pass_reset_url` + `user_cancel_url` (both `->toString()`) |
| simple_pass_reset | `user_pass_rehash` (nested in `hash_equals`) |

## Quality gates

- `composer fix-style` — clean
- `composer phpstan` — no errors
- PHPUnit — 3/3 (above-version transform, below-version no-op, unrelated-function no-change)
- rector-qa — all 5 passes green (type-guard SAFE, fixtures PASS, BC PASS, @see PASS, registration PASS)

## Notes

A `TODO PHPSTAN_MESSAGES` block is recorded in the config: the test core (11.4-dev) predates commit `a38f1d17`, so PHPStan does not yet emit the deprecation for these symbols. The three messages should be captured once the test core advances past that commit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,20 @@ release-by-release.
 
 ## [Unreleased]
 
+### Added
+
+- **`ReplaceUserOneTimeAuthFunctionsRector` (Drupal 11.4, [#3581056](https://www.drupal.org/node/3581056))** —
+  replaces the deprecated user one-time authentication functions
+  `user_pass_rehash()`, `user_pass_reset_url()` and `user_cancel_url()` (deprecated
+  in drupal:11.4.0, removed in drupal:13.0.0) with the new
+  `\Drupal\user\OneTimeAuthentication` service methods `generateHmac()`,
+  `generateOneTimeLoginUrl()` and `generateCancelConfirmUrl()`. BC-wrapped via
+  `DeprecationHelper` because the service does not exist on Drupal < 11.4. The two
+  URL methods return a `Url` object, so the rewrite chains `->toString()`.
+  `user_mail_tokens()` (deprecated in the same issue) is intentionally not handled —
+  it is used as a string callback and its replacement needs a `BubbleableMetadata`
+  argument that cannot be synthesised.
+
 ## [1.0.0-beta1] — 2026-06-11
 
 ### Added
--- a/config/drupal-11/drupal-11.4-deprecations.php
+++ b/config/drupal-11/drupal-11.4-deprecations.php
@@ -33,6 +33,7 @@ use DrupalRector\Drupal11\Rector\Deprecation\ReplaceNonBoolAccessRector;
 use DrupalRector\Drupal11\Rector\Deprecation\ReplaceRecipeRunnerInstallModuleRector;
 use DrupalRector\Drupal11\Rector\Deprecation\ReplaceSessionManagerDeleteRector;
 use DrupalRector\Drupal11\Rector\Deprecation\ReplaceSystemPerformanceGzipKeyRector;
+use DrupalRector\Drupal11\Rector\Deprecation\ReplaceUserOneTimeAuthFunctionsRector;
 use DrupalRector\Drupal11\Rector\Deprecation\ReplaceViewsProceduralFunctionsRector;
 use DrupalRector\Drupal11\Rector\Deprecation\SystemRegionFunctionsRector;
 use DrupalRector\Drupal11\Rector\Deprecation\SystemSortThemesRector;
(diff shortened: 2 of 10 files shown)


bbrala:
## Review: `ReplaceUserOneTimeAuthFunctionsRector` — ready to merge ✅

Verified the rewrite against the **real** upstream `11.x` core (not just the local mirror, which is stale at Jun 4).

### Correctness — verified against upstream 11.x
The deprecation landed in real `11.x` (tip `4b89f8b5`, Jun 17) and was **not reverted** — `core/modules/user/src/OneTimeAuthentication.php` exists and `user.module` carries the `@deprecated` annotations + service delegations. The rector's mappings exactly mirror core's own delegation, including the service method signatures:

| Function | Service method (real signature) | Rewrite |
|---|---|---|
| `user_pass_rehash($account, $timestamp)` | `generateHmac(UserInterface $account, int $timestamp): string` | ✓ |
| `user_pass_reset_url($account, $options)` | `generateOneTimeLoginUrl(UserInterface $account, array $options = [], bool $immediate = FALSE): Url` | ✓ `->toString()` |
| `user_cancel_url($account, $options)` | `generateCancelConfirmUrl(UserInterface $account, array $options = []): Url` | ✓ `->toString()` |

- BC-wrapping is justified — the service does not exist pre-11.4.
- `user_mail_tokens()` exclusion is correct: its replacement `OneTimeAuthentication::tokens()` requires a `BubbleableMetadata` argument that can't be synthesised, and it's used as a string callback.

> Minor note: the PR body cites commit `a38f1d17`, which isn't reachable in upstream history because the MR was rebased/squashed on merge — the substance landed regardless. Worth noting the citation is the pre-merge commit.

### Conventions — ✅
`configure()` override + `protected array $configuration` redeclaration + `DrupalIntroducedVersionConfiguration` validation matches the pattern used by all 52 sibling rectors. Exact global-function name matching; the `no_change_unrelated` fixture proves similarly-named functions are left alone.

### Test coverage — strengthened
The committed fixtures originally covered only top-level statement calls. The dominant real-world contrib shape nests `user_pass_rehash()` inside `hash_equals()` — confirmed in `tfa` (`src/EventSubscriber/TfaOneTimeLoginEventSubscriber.php:126`), `change_pwd_page`, and `simple_pass_reset`. Added a `nested_hash_equals.php.inc` fixture (commit `2d82ee62`) verifying the inner call is rewritten in place and BC-wrapped while `hash_equals()` stays intact. Suite now 4/4 green.

**Verdict:** correct, follows conventions, verified against real upstream core. Good to merge.

PR #367, opened 2026-06-12
feat(11.4): add NodeSearch → SearchNode rename to DRUPAL_114_BREAKING (#3587564)

bbrala:
## Summary

Adds the `NodeSearch` → `SearchNode` class rename to the opt-in `DRUPAL_114_BREAKING` set.

`Drupal\node\Plugin\Search\NodeSearch` was moved out of the `node` module into the new `search_node` core sub-module and renamed to `Drupal\search_node\Plugin\Search\SearchNode` in **drupal:11.4.0** ([#3587564](https://www.drupal.org/i/3587564) / [change record #3590298](https://www.drupal.org/node/3590298)). Registered as a `RenameClassRector` mapping alongside the existing `HelpSearch` → `SearchHelpSearch` entry.

## Why the breaking set (not the default deprecation set)

- `SearchNode` does **not** exist on any Drupal minor below 11.4, so rewriting `use` / `extends` / `::class` references to it would produce a "class not found" fatal on < 11.4.
- A `class X extends Y` declaration is a structural (Name-node) change, not an `Expr → Expr` rewrite, so it **cannot** be BC-wrapped via `DeprecationHelper`.

This is the same shape as the `HelpSearch` precedent already in this file. One difference, documented inline: unlike HelpSearch (moved away with no shim), `NodeSearch` survives in 11.4 as a `@deprecated` subclass of `SearchNode` (removed in 12.0.0), so PHPStan *does* emit a deprecation message for it.

## Verification

- `php -l`, `composer fix-style` (0 changes), `composer phpstan` (no errors) — all clean.
- **Live test** (`rector-live-test`) against real D11 contrib confirmed correct transformation:
  - `trash` (~19k installs) — `TrashNodeSearch extends NodeSearch` → `extends \Drupal\search_node\Plugin\Search\SearchNode`
  - `search_exclude` (~4.7k installs) — `SearchExcludeNodeSearch` likewise
- PHPStan message capture was not possible in the test env (its `11.4-dev` core snapshot predates the move); the expected message form is documented as a prose comment, intentionally outside the coverage-registry capture format (matching the HelpSearch/NodeViewController precedent).

## Files

- `config/drupal-11/drupal-11.4-breaking.php` — `RenameClassRector` mapping + documenting comment
- `docs/implemented-digests.yml` — recorded `#3587564` as `config-only`
- `CHANGELOG.md` — `[Unreleased]` entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,24 @@ release-by-release.
 
 ## [Unreleased]
 
+### Added
+
+- **`DRUPAL_114_BREAKING`: `NodeSearch` → `SearchNode` class rename.**
+  `Drupal\node\Plugin\Search\NodeSearch` was moved out of the `node` module and
+  renamed to `Drupal\search_node\Plugin\Search\SearchNode` in the new
+  `search_node` core sub-module (drupal:11.4.0). Added as a `RenameClassRector`
+  entry to `drupal-11.4-breaking.php`. Unlike the `HelpSearch` move, the old
+  `NodeSearch` class is **not** removed in 11.4 — it survives as a `@deprecated`
+  subclass of `SearchNode` until removal in drupal:12.0.0. It is still in the
+  breaking set because `SearchNode` does not exist on any Drupal minor below
+  11.4, and a `use` / `extends` / `::class` rename is structural and cannot be
+  BC-wrapped, so applying it against code that must still run on an older minor
+  would fatal there. Known D11 contrib subclasses affected: `trash`
+  (`TrashNodeSearch`) and `search_exclude` (`SearchExcludeNodeSearch`). Opt in
+  via `Drupal11SetList::DRUPAL_114_BREAKING` only after dropping support for
+  Drupal < 11.4 ([#3587564](https://www.drupal.org/i/3587564) /
+  [change record](https://www.drupal.org/node/3590298)).
+
 ## [1.0.0-beta1] — 2026-06-11
 
 ### Added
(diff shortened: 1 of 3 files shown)


bbrala:
## Review: clean — approve ✅

Verified every claim against drupal-core (\`f53b82e40c\`).

### Correctness — all verified against core
- **Target exists:** \`core/modules/search/modules/search_node/src/Plugin/Search/SearchNode.php\` — namespace + FQCN match the mapping exactly.
- **Old class is a deprecated shim:** \`NodeSearch\` is \`@deprecated in drupal:11.4.0 and is removed from drupal:12.0.0. Instead, use \Drupal\search_node\Plugin\Search\SearchNode.\` — matches the CHANGELOG/comment (12.0.0 removal) precisely.
- **CR matches:** core shim's \`@see\` is \`3590298\`, exactly as cited.
- **Digest fidelity:** the digest file uses the identical \`RenameClassRector\` mapping → \`config-only\` is the correct classification.
- **Breaking-set placement is sound:** \`SearchNode\` does not exist below 11.4, and a \`use\`/\`extends\`/\`::class\` rename is structural (not \`Expr→Expr\`), so it can't be \`DeprecationHelper\`-wrapped. Same shape as the \`HelpSearch\` precedent.

### Pure missing-symbol break, not behavioral
Unlike \`NodeViewController\`, this is **purely** a missing-symbol break. The \`NodeSearch\` shim (49 lines) declares only \`__construct\` (delegate + \`@trigger_error\`) and **no other overrides**, so reparenting a subclass onto \`SearchNode\` is behaviorally identical. The inline comment correctly claims only the missing-symbol + structural reasons and does **not** overclaim a behavioral break — accurate.

### Notes
- The \`PHPSTAN_MESSAGES\` block being prose-only is correct: \`RenameClassRector\` is a generic rule with no bespoke class to hold a const, matching the file's convention.
- Live-test PHPStan capture being impossible (test-env core predates the move) is documented honestly; the transformation was validated on \`trash\` and \`search_exclude\`.

No correctness, convention, BC, or coverage issues found.

PR #368, opened 2026-06-13
feat: add UserLoadByNameAndMailRector for #3555936

bbrala:
## Summary

Adds `UserLoadByNameAndMailRector`, covering the two functions deprecated by change record [#3555936](https://www.drupal.org/node/3555936):

- `user_load_by_name()` — deprecated in drupal:11.4.0, removed in drupal:13.0.0
- `user_load_by_mail()` — deprecated in drupal:11.4.0, removed in drupal:13.0.0

No upstream drupal-digests rule existed for this, so the rector is authored from scratch.

## Transformation

```php
// before
$account = user_load_by_name($name);

// after
$account = array_values(\Drupal::entityTypeManager()->getStorage('user')->loadByProperties(['name' => $name]))[0] ?? FALSE;
```

`loadByProperties()` returns an array keyed by user ID, whereas the old functions returned a single user object or `FALSE`. The result is normalised with `array_values(...)[0] ?? FALSE` to preserve the original single-object-or-`FALSE` contract in any call context (assignment, `if (!…)`, argument), without an intermediate variable and without passing an rvalue to `reset()`.

## BC decision

`AbstractRector` (no `DeprecationHelper` wrapper). The replacement uses only long-standing entity APIs (`entityTypeManager()`, `getStorage()`, `loadByProperties()`) plus pure PHP. Verified against the `10.1.0` core tag: all three symbols exist with identical signatures, `loadByProperties()` already returns an id-keyed array, and `user_load_by_name()` itself is implemented with the exact same lookup — so the output runs unchanged on every supported Drupal version.

## Validation

- Unit test passes (`testAboveVersion`-style happy path + `no_change_unrelated` fixture).
- `phpstan` clean, `fix-style` no changes.
- **Live-tested** against 6 D11 contrib modules — 13 call sites across `.module`, `.install`, Controller, Form, FieldFormatter, and Plugin contexts (login_security, login_emailusername, linkchecker, workbench_access, restrict_password_change, show_email). Operator precedence verified correct (`!(… ?? FALSE)` parenthesized; `if ($x = … ?? FALSE)` left bare).
- Real PHPStan deprecation messages for both functions captured in `docs/coverage-registry.php`.

## Notes

- Registered in the `DRUPAL_114` deprecation set.
- Recorded in `docs/implemented-digests.yml` (#3555936) and `CHANGELOG.md`.
- `docs/coverage-registry.php` is regenerated; besides this rector's two entries it also syncs previously-unregenerated messages for already-merged rectors whose source consts were already on `main`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,19 @@ release-by-release.
 
 ## [Unreleased]
 
+### Added
+
+- **`UserLoadByNameAndMailRector` (Drupal 11.4, [#3555936](https://www.drupal.org/node/3555936))** —
+  rewrites the deprecated `user_load_by_name()` and `user_load_by_mail()`
+  functions (deprecated in drupal:11.4.0, removed in drupal:13.0.0) to the
+  equivalent entity-storage lookup
+  `\Drupal::entityTypeManager()->getStorage('user')->loadByProperties([...])`.
+  Because `loadByProperties()` returns an array keyed by user ID while the old
+  functions returned a single user object or `FALSE`, the result is normalised
+  with `array_values(...)[0] ?? FALSE` to preserve the original return contract.
+  The replacement uses only long-standing entity APIs and pure PHP, so it runs
+  on every supported Drupal version and is not BC-wrapped.
+
 ## [1.0.0-beta1] — 2026-06-11
 
 ### Added
--- a/config/drupal-11/drupal-11.4-deprecations.php
+++ b/config/drupal-11/drupal-11.4-deprecations.php
@@ -38,6 +38,7 @@ use DrupalRector\Drupal11\Rector\Deprecation\SystemRegionFunctionsRector;
 use DrupalRector\Drupal11\Rector\Deprecation\SystemSortThemesRector;
 use DrupalRector\Drupal11\Rector\Deprecation\UploadedFileConstraintArrayOptionsToNamedArgsRector;
 use DrupalRector\Drupal11\Rector\Deprecation\UseEntityTypeHasIntegerIdRector;
+use DrupalRector\Drupal11\Rector\Deprecation\UserLoadByNameAndMailRector;
 use DrupalRector\Drupal11\Rector\Deprecation\ViewsPluginHandlerManagerRector;
 use DrupalRector\Rector\Deprecation\ClassConstantToClassConstantRector;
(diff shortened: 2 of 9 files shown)


bbrala:
## Review — Approve ✅

Reviewed correctness, BC decision, false-positive risk, and the return-contract preservation. The implementation is solid.

### Correctness
- **Matches core semantics.** The deprecated functions internally do `loadByProperties(['name'|'mail' => …])` then `reset()`. `array_values(...)[0]` returns the same first element from the uid-keyed array; `array_values` is genuinely needed since the array isn't 0-keyed.
- **Empty-result contract preserved.** No match → `array_values([])[0] ?? FALSE` → `FALSE`, and `??` correctly suppresses the "undefined key 0" warning. This is exactly why `?? FALSE` beats `reset()` here.
- **No false positives.** `getNodeTypes()` is `FuncCall` only, so `$this->user_load_by_name()` (MethodCall) and a namespaced `Foo\user_load_by_name()` won't match. The `getName()` + `isset(PROPERTY_MAP[...])` gate is correct.
- **Operator precedence** is delegated to php-parser's pretty-printer (returns a `Coalesce` node), which parenthesizes by AST precedence.

### BC decision — correct
Output uses only long-standing entity APIs (`entityTypeManager()`/`getStorage()`/`loadByProperties()`) + pure PHP → `AbstractRector`, no `DeprecationHelper`. Verified against the 10.1.0 tag.

### One gap, now closed
The fixtures covered plain assignment + a no-change case, but not the precedence-sensitive contexts the PR's risk hinges on. Following the repo convention (`PluginBaseIsConfigurableRector/fixture/negated.php.inc`, `ReplaceUserOneTimeAuthFunctionsRector/fixture/nested_hash_equals.php.inc`), I added a `negated.php.inc` fixture covering:

- `if (!user_load_by_name($name))` → `if (!(array_values(…)[0] ?? FALSE))` — negation parenthesizes the `??`
- `do_something(user_load_by_mail($mail))` → argument context, left bare

The rector produced exactly the expected output on first run, locking in the parenthesization. Suite passes: `OK (3 tests, 4 assertions)`.

(Minor, not blocking: the `docs/coverage-registry.php` regeneration adds trailing whitespace on 7 lines — a `var_export()` artifact.)

PR #370, opened 2026-06-16
feat: composer-based sets + register as rector-extension

bbrala:
## What

Two related changes that let Rector configure Drupal deprecation fixes from the installed `drupal/core` version.

### 1. Composer-based sets (`DrupalSetProvider`)
A `SetProviderInterface` exposing the per-minor deprecation **and** breaking configs as `ComposerTriggeredSet`s keyed on `drupal/core`. Enable with:

```php
return RectorConfig::configure()
    ->withSetProviders(\DrupalRector\Set\DrupalSetProvider::class)
    ->withComposerBased(drupal: true);
```

- Cumulative-downward via `^version`: a site on 11.4 loads 11.0→11.4 and **never** a future minor — inherently BC-safe (only rules for deprecations that are live on the installed core). Verified incl. dev/beta core.
- The exact-version match makes the **breaking** sets safe to fold into the same `drupal` group.
- The PHPUnit bootstrap is supplied by a dedicated `config/drupal-bootstrap.php` set, matched once per major (`drupal/core ^10.0` / `^11.0`). The per-minor deprecation configs don't register it (only the aggregated `*-all-deprecations.php` sets do), and composer-based selection loads the per-minor configs directly — so the bootstrap is registered separately. Kept apart on purpose: the bootstrap throws without a Drupal install, and composer-based sets only ever load when `drupal/core` is present, so the throw can never fire here.
- Covers `drupal/core` 10.0–10.3 and 11.0–11.4.

### 2. Register as a `rector-extension` (progressive enhancement)
`type: rector-extension` + `extra.rector.includes` ships `config/extension.php`, which supplies sensible Drupal defaults (file extensions, autoloading, the `upgrade_status` skip, phpstan-drupal). It loads **before** the project's `rector.php`, so the project config always wins.

**The bundled `rector.php` template is left unchanged / self-sufficient on purpose** — see the testing note below.

## Why the template wasn't slimmed (empirical finding)

I tested whether the auto-config loads in a real consumer (scratch project, sentinel in `config/extension.php`, ran `rector --dry-run`):

| Consumer setup | `extension.php` auto-loaded? |
|---|---|
| `rector/rector` + `drupal-rector` | ❌ No |
| + `rector/extension-installer` | ✅ Yes |

So the auto-config **only** activates when the consumer has `rector/extension-installer` (`rector/rector` doesn't pull it in, and Composer silently skips plugins not in `allow-plugins`). Slimming the template would therefore risk silently dropping `.module`/`.theme` coverage for projects without the installer. The template stays the reliable source of truth; the extension is a bonus for projects that have the installer.

The same test caught a bug: `DrupalFinderComposerRuntime::getDrupalRoot()` **throws** (`Package "drupal/core" is not installed`) rather than returning null, so the lookup is now guarded with `InstalledVersions::isInstalled('drupal/core')` and re-verified to not throw on non-Drupal projects.

## Upstream dependency
`withComposerBased(drupal: true)` needs Rector core to ship `SetGroup::DRUPAL` + the `drupal:` toggle — rectorphp/rector-src#8041. Until then the provider is dormant. Plan per @TomasVotruba: get the provider into `main`, then kick off testing upstream.

## Verification
- Full PHPStan clean, full unit suite green (659 tests), `composer validate` OK.
- `DrupalSetProviderTest`: 8 tests / 82 assertions incl. the cumulative-matching matrix (deprecation + breaking + bootstrap) and the no-`drupal/core` case.
- Auto-config load behaviour + the non-Drupal no-throw path verified by hand in a scratch consumer project.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,8 @@ release-by-release.
   It loads before your `rector.php`, so your own config always takes precedence,
   and the bundled `rector.php` template remains fully self-sufficient when the
   installer is absent. The PHPUnit bootstrap is intentionally not auto-loaded
-  (it requires a Drupal install) and stays tied to the deprecation sets.
+  (it requires a Drupal install) — under composer-based selection it is loaded by
+  a dedicated `config/drupal-bootstrap.php` set instead.
 - **Composer-based sets ([rectorphp/rector#9778](https://github.com/rectorphp/rector/issues/9778))** —
   new `DrupalRector\Set\DrupalSetProvider` lets Rector select Drupal deprecation
   sets automatically from the installed `drupal/core` version, via
@@ -31,8 +32,10 @@ release-by-release.
   Matching is cumulative downward (a site on 11.4 loads the 11.0 → 11.4 sets and
   never a future minor), which is backward-compatibility safe. Because the
   installed version is known exactly, the per-minor *breaking* sets are folded in
-  and applied automatically. Covers `drupal/core` 10.0–10.3 and 11.0–11.4.
-  Requires a Rector release that ships `SetGroup::DRUPAL` and the
+  and applied automatically. The PHPUnit bootstrap (which the per-minor configs
+  don't register) is supplied by a dedicated `config/drupal-bootstrap.php` set,
+  matched once per major (`^10.0` / `^11.0`). Covers `drupal/core` 10.0–10.3 and
+  11.0–11.4. Requires a Rector release that ships `SetGroup::DRUPAL` and the
   `withComposerBased(drupal: ...)` toggle.
 
 ## [1.0.0-beta1] — 2026-06-11
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,29 @@ release-by-release.
 
 ## [Unreleased]
 
+### Added
+
+- **drupal-rector now registers as a Rector extension** — the package `type` is
+  `rector-extension` and it ships `config/extension.php` (via
+  `extra.rector.includes`). For projects that have `rector/extension-installer`,
+  Rector auto-loads it to supply sensible Drupal defaults — the extra PHP file
+  extensions (`module`, `theme`, `install`, …), class-import conventions, the
+  `upgrade_status` skip, Drupal autoloading and (when present) phpstan-drupal.
+  It loads before your `rector.php`, so your own config always takes precedence,
+  and the bundled `rector.php` template remains fully self-sufficient when the
+  installer is absent. The PHPUnit bootstrap is intentionally not auto-loaded
+  (it requires a Drupal install) and stays tied to the deprecation sets.
+- **Composer-based sets ([rectorphp/rector#9778](https://github.com/rectorphp/rector/issues/9778))** —
+  new `DrupalRector\Set\DrupalSetProvider` lets Rector select Drupal deprecation
+  sets automatically from the installed `drupal/core` version, via
+  `->withSetProviders(DrupalSetProvider::class)->withComposerBased(drupal: true)`.
+  Matching is cumulative downward (a site on 11.4 loads the 11.0 → 11.4 sets and
+  never a future minor), which is backward-compatibility safe. Because the
+  installed version is known exactly, the per-minor *breaking* sets are folded in
+  and applied automatically. Covers `drupal/core` 10.0–10.3 and 11.0–11.4.
+  Requires a Rector release that ships `SetGroup::DRUPAL` and the
+  `withComposerBased(drupal: ...)` toggle.
+
 ## [1.0.0-beta1] — 2026-06-11
 
 ### Added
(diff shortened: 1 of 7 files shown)


bbrala:
Initial exploration, some issues need examination, especially the rector.php, but think this is 90%

config/extension.php
```diff
@@ -0,0 +1,71 @@
...
+ * deprecation sets instead (see \DrupalRector\Set\DrupalSetProvider and the
+ * *-all-deprecations sets), which only run against an actual Drupal codebase.
+ */
+return static function (RectorConfig $rectorConfig): void {
+    // Drupal executes PHP from several non-.php extensions.
+    $rectorConfig->fileExtensions(['php', 'module', 'theme', 'install', 'profile', 'inc', 'engine']);
+
+    // Drupal convention: import class names, but leave docblock names untouched.
+    $rectorConfig->importNames(true, false);
+    $rectorConfig->importShortClasses(false);
```
TomasVotruba: This might be convention to Drupal, but we typically leave this up to end-user. Some project prefer different approach and it might a bit surprising this settings change once they instill this package.

config/extension.php
```diff
@@ -0,0 +1,71 @@
...
+
+    $drupalRoot = $drupalFinder->getDrupalRoot();
+    if (is_string($drupalRoot) && $drupalRoot !== '') {
+        $rectorConfig->autoloadPaths([
+            $drupalRoot.'/core',
+            $drupalRoot.'/modules',
+            $drupalRoot.'/profiles',
+            $drupalRoot.'/themes',
+        ]);
+    }
```
TomasVotruba, line 57: Looks good 👍

CHANGELOG.md
```diff
@@ -12,6 +12,32 @@ release-by-release.
...
+  sets automatically from the installed `drupal/core` version, via
+  `->withSetProviders(DrupalSetProvider::class)->withComposerBased(drupal: true)`.
+  Matching is cumulative downward (a site on 11.4 loads the 11.0 → 11.4 sets and
+  never a future minor), which is backward-compatibility safe. Because the
+  installed version is known exactly, the per-minor *breaking* sets are folded in
+  and applied automatically. The PHPUnit bootstrap (which the per-minor configs
+  don't register) is supplied by a dedicated `config/drupal-bootstrap.php` set,
+  matched once per major (`^10.0` / `^11.0`). Covers `drupal/core` 10.0–10.3 and
+  11.0–11.4. Requires a Rector release that ships `SetGroup::DRUPAL` and the
+  `withComposerBased(drupal: ...)` toggle.
```
TomasVotruba: This is a bit hard to read. Maybe Rector config example as in PR description would be nice 👍

src/Set/DrupalSetProvider.php
```diff
@@ -0,0 +1,108 @@
...
+ * installed core, which is inherently backward-compatibility safe.
+ *
+ * Because the matched version guarantees the replacement symbols exist on the
+ * installed core, the otherwise opt-in "breaking" sets are safe to include here
+ * and are folded into the same group.
+ *
+ * Enable in rector.php:
+ *
+ *     return RectorConfig::configure()
+ *         ->withSetProviders(\DrupalRector\Set\DrupalSetProvider::class)
```
TomasVotruba: This line is typically installed via the extension autoregistration. Users should not handle this manually.

Also, this allows to change or have mulptiple sets providers later on and structure inner sets differently. (e.g. drupal type coverage set provider etc.), without bothering user to register them.

In short, `DrupalSetProvider` is internal class, not for the user.

src/Set/DrupalSetProvider.php
```diff
@@ -0,0 +1,108 @@
...
+ *         ->withComposerBased(drupal: true);
```
TomasVotruba: This is the line for end-user, to enable willingly if then want to turn on Drupal upgrade sets 👍

tests/src/Set/DrupalSetProviderTest.php
```diff
@@ -0,0 +1,128 @@
...
+                $matched[] = basename($set->getSetFilePath());
+            }
+        }
+
+        sort($matched);
+        $expected = $expectedFiles;
+        sort($expected);
+
+        self::assertSame($expected, $matched);
+    }
```
TomasVotruba, line 120: Cool test 👍

TomasVotruba:
I've added few comments. Looks very good to me  :+1:

config/extension.php
bbrala: Hmm, yeah i think i agree, i can add it to the default rector.php added in the repository, but not load it as is when selecting sets.

CHANGELOG.md
bbrala: Yeah you're right. We'll simplify to something more sane xD

src/Set/DrupalSetProvider.php
bbrala: Hmm, ok. Can you please confirm the autoloading of the extension is automaticly done always? It was a. bit opague if that is indeed the case.

src/Set/DrupalSetProvider.php
bbrala: This reads like its not the same as the comment above? The class is internal and users shouldn't use it, but this is also the code the users need to use? I'm a little confused, sorry :)

src/Set/DrupalSetProvider.php
bbrala: Ah, comment versus code. Never mind me.

src/Set/DrupalSetProvider.php
bbrala: Ok, think it has cleared up ;)

PR #398, opened 2026-06-24
fix: capture by-reference arguments in BC-call closures

bbrala:
## Problem

`DeprecationHelper::backwardsCompatibleCall()` invokes its callables with **no arguments**, so the converted call has to close over the caller's arguments. `AbstractDrupalCoreRector::createBcCallOnExpr()` used **arrow functions** for this — and arrow functions capture **by value**. Any mutation made through a **by-reference parameter** was therefore silently dropped on the rewritten code.

This breaks replacements such as:

- `template_preprocess_*(&$variables)` → `ThemePreprocess::preprocess*()` etc.
- form `*_process` / `*_validate` / `*_builder` callbacks taking `&$form`
- `views_add_contextual_links(&$render_element)`

In webform it surfaced as failing render/form PHPUnit tests on the generated patch (e.g. `Undefined array key "html_attributes"`), because the `&$variables` populated inside `template_preprocess_html()` never made it back to the caller.

## Fix

`createBcCallOnExpr()` now detects by-reference parameters and, when a call argument is a local variable bound to one, emits a **long closure that captures it by reference** instead of an arrow function:

```php
DeprecationHelper::backwardsCompatibleCall(\Drupal::VERSION, '11.2.0',
    function () use (&$variables) { \Drupal::service(ThemePreprocess::class)->preprocessHtml($variables); },
    function () use (&$variables) { template_preprocess_html($variables); },
);
```

By-reference parameters are discovered by **reflecting the replacement service method** (and, best-effort, the deprecated function), so detection is automatic for every affected deprecation — no per-rule flags. When no argument is bound to a by-reference parameter, the cleaner arrow function is kept.

### void vs. return

PHPStan flags `return <void-expr>;` as `function.void` at **every** level, while allowing both an arrow function's implicit return and a bare expression statement. So a closure wrapping a **void** target emits a bare `<expr>;` statement, and only a value-returning target keeps `return <expr>;`. The replacement method's reflected return type decides which form is emitted. `DeprecationHelperRemoveRector` unwraps both forms.

## Changes

- `AbstractDrupalCoreRector::createBcCallOnExpr()` — by-reference detection + long-closure emission, void-aware closure body.
- `DeprecationHelperRemoveRector` — unwraps both `return` and bare-expression closures.
- Classmap stubs for the reflected target classes (correct `&` positions + return types) so detection is deterministic in the test suite.
- Fixtures covering preprocess, content_translation, menu_ui, media_library and views by-reference conversions.
- `rector-implement` QG-C and `rector-qa` Pass 7 document a by-reference audit step for future rules.

## Testing

Full suite green (694 tests). Source PHPStan clean.

Companion issue: https://git.drupalcode.org/project/rector/-/work_items/3600785

### DIFF
--- a/.claude/skills/rector-implement/SKILL.md
+++ b/.claude/skills/rector-implement/SKILL.md
@@ -181,6 +181,50 @@ transformation.
 
 ---
 
+### After QG-B: Quality Gate QG-C — By-Reference Parameter Audit (BC-wrapped rectors only)
+
+Apply only if the rector extends `AbstractDrupalCoreRector` and the conversion happens in an
+**expression context** (the call gets wrapped in `DeprecationHelper::backwardsCompatibleCall`).
+This catches the class of bug fixed in `3ac6e255`: the BC wrapper closes over the caller's
+arguments, and an arrow function captures **by value**, so any mutation made through a
+**by-reference parameter** (`&$variables`, `&$form`, `&$element`, …) is silently dropped on the
+rewritten code — breaking the module at runtime (e.g. `template_preprocess_html(&$variables)`).
+
+1. **Check both signatures for `&` parameters.** Look at the replacement target method (the
+   class/method in your `FunctionToServiceConfiguration`) **and** the deprecated function, in
+   `repos/drupal-core`:
+   ```bash
+   rg -n 'function preprocessHtml|function template_preprocess_html' repos/drupal-core/core
+   ```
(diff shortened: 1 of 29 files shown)


bbrala:
Tested here: https://git.drupalcode.org/project/project_analysis/-/pipelines/863334

All issues with webform seem fixed.

PR #419, opened 2026-08-10
feat: add composer-based set binding every rule to an exact drupal/core version

TomasVotruba:
Hey, I'm improving the way Rector find composer-based rules and sets. All you need is:

* 1 config with all the rules for all the version
* make rule implement `ComposerPackageConstraintInterface`
* or use  `ruleWithConfigurationComposerVersionBound` - e.g. for generic rules like rename class:
 
```php
  $rectorConfig->ruleWithConfigurationComposerVersionBound(
            RenameClassRector::class,
            [
                'SomeOldClass' => 'SomeNewClass',
            ],
            'phpunit/phpunit',
            '>=9.0'
        );
```
 

The set provider, SetList, versioned set list can be dropped (in one of next PRs, not now).

It's agent generated, but it works well on Doctrine/Symfony/PHPUnit sets, see e.g.

* https://github.com/rectorphp/rector-doctrine/pull/497
* https://github.com/rectorphp/rector-doctrine/pull/501
* https://github.com/rectorphp/rector-doctrine/pull/502

I'll open this for feedback if it's all done correctly :) do not merge yet.
I'll make sure it's part of next Rector release and everything works as smoothly as before.

Feedback welcomed :+1: 

I've also added custom PHPStan rule to make sure all the rules are registered in `composer-based.php` config and nothing is missed by accident :wink: 

---

Replaces `DrupalSetProvider` — Rector deprecated `SetProviderInterface` and `ComposerTriggeredSet` in favour of binding rules to a package version directly - with a single generated set, `config/composer-based.php`, exposed as `DrupalSetList::COMPOSER_BASED`.


TomasVotruba:
Still WIP... need a bit config tuning. But check the linked PRs in the description to get the idea ahead :+1:

TomasVotruba:
Ready now :+1:

TomasVotruba:
Ping @bbrala

tobiasbaehr:
We should bump the required rector version and make `\DrupalRector\Set\DrupalSetProvider` a no-op class, because it tries to call constant which was removed in rector.

So that ` ->withSetProviders(DrupalSetProvider::class)` does nothing, until rector removes also `withSetProviders` method.

Or all sets needs to check that the constant like `PHPUnitSetList::PHPUNIT_90 ` are defined, but without to throw an exception when not, like what it does in drupal-10.0-deprecations.php.

bbrala:
Ok, back from vacation. I did a ninja release for this package to lock it out of >2.6.1 so it doesnt break while this is worked on. 

There is another path in this, and that is removing the specific rules for twig and phpunit maybe. Also kinda interested in how this works.

A rule tells us: ">=10.1.0", does that mean it is selected on anything above, or anything above but belore next major?

So does `>=10.1.0` equal `>=10.1.0 <= 10.999.999` or just `>=10.1.0`.

I wouldn't mind to much to drop the set references, but am a little worried to throw things around again with how the sets work.

TomasVotruba:
Welcome back :) many news in Rector.

Just the >=. 
If it changes in 11 differently, it should be limited from upper bound: < 11

Checkout my PRs in Laravel Rector package, where I did the same upgrade.

bbrala:
I think i will do a min/max for the contraints. This does mean some archeoligy to when things actually god removed, but i think that should work. 

I really really don't want all rules to run in every major, but probably only the rules of the current major (very very mayby also the previous). This seems like a way to keep stability in check and not just create a drag of old rules that need to work on newer versions.

